PR #3, opened 2021-04-21
Apl soc fixes

miczyg1:


miczyg1:
This fixes `i801_smbus 0000:00:1f.1: can't derive routing for PCI INT A` and `x86/cpu: VMX (outside TXT) disabled by BIOS` problems seen in dmesg.

PR #5, opened 2021-05-03
WIP: DWS v0.2-rc4 and initial support for custom SMBIOS BIOS Vendor

pietrushnic:
Signed-off-by: Piotr Król <piotr.krol@3mdeb.com>


src/arch/x86/smbios.c
```diff
@@ -408,7 +408,7 @@ static int smbios_write_type0(unsigned long *current, int handle)
 	t->handle = handle;
 	t->length = len - 2;
 
-	t->vendor = smbios_add_string(t->eos, "coreboot");
+	t->vendor = smbios_add_string(t->eos, CONFIG_MAINBOARD_SMBIOS_VENDOR);
```
miczyg1, line 411:
```suggestion
	t->vendor = smbios_add_string(t->eos, smbios_mainboard_bios_vendor());
```
otherwise what for we create such function in `src/arch/x86/smbios_defaults.c`?

pietrushnic: @miczyg1 yeah this is not finished, that's why there is no such call here.

PR #21, opened 2021-11-02
cpu/power/spr.h: change SPR numbers definitions to decimal

krystian-hebel:
ISA lists those values in decimal, this change makes it easier to
compare with documentation. It also fixes wrong values for SRR0/SRR1.

Signed-off-by: Krystian Hebel <krystian.hebel@3mdeb.com>
Change-Id: I326f9ec73ab63c4c1cef836d3217e7132f49a4f5

### DIFF
--- a/src/include/cpu/power/spr.h
+++ b/src/include/cpu/power/spr.h
@@ -3,23 +3,23 @@
 
 #include <arch/byteorder.h>	// PPC_BIT()
 
-#define SPR_DEC					0x16
+#define SPR_DEC					22
 #define SPR_DEC_IMPLEMENTED_BITS		56
 #define SPR_DEC_LONGEST_TIME			((1ull << (SPR_DEC_IMPLEMENTED_BITS - 1)) - 1)
 
-#define SPR_SRR0				0x2A
-#define SPR_SRR1				0x2B
+#define SPR_SRR0				26
+#define SPR_SRR1				27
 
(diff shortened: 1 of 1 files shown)


src/include/cpu/power/spr.h
```diff
@@ -3,23 +3,23 @@
...
 
-#define SPR_DEC					0x16
+#define SPR_DEC					22
 #define SPR_DEC_IMPLEMENTED_BITS		56
 #define SPR_DEC_LONGEST_TIME			((1ull << (SPR_DEC_IMPLEMENTED_BITS - 1)) - 1)
 
-#define SPR_SRR0				0x2A
-#define SPR_SRR1				0x2B
+#define SPR_SRR0				26
+#define SPR_SRR1				27
```
SergiiDmytruk, line 11:
```suggestion
#define SPR_SRR0				42
#define SPR_SRR1				43
```

SergiiDmytruk (changes requested):
Two values are off by 16.

src/include/cpu/power/spr.h
krystian-hebel: That's the issue, they were wrong previously (my bad). 26 and 27 are correct values: https://wiki.raptorcs.com/w/images/c/cb/PowerISA_public.v3.0B.pdf page 971

SergiiDmytruk (approved):

PR #22, opened 2021-11-02
Setup watchdog in romstage for 2 minutes

SergiiDmytruk:
Fixes Dasharo/dasharo-issues#29.

Also handles IPMI issues discovered in the process.


src/soc/ibm/power9/romstage.c
```diff
@@ -340,6 +342,16 @@ void main(void)
...
+	if (ipmi_premem_init(CONFIG_BMC_BT_BASE, 0) != CB_SUCCESS)
+		die("Failed to initialize IPMI\n");
+
+	/*
+	 * Two minutes to load.
+	 * Not handling return code, because the function itself prints log messages
+	 * and its failure is not a critical error.
+	 */
+	(void)ipmi_init_and_start_bmc_wdt(CONFIG_BMC_BT_BASE, 120, TIMEOUT_HARD_RESET);
+
```
krystian-hebel, line 354: Please move it below console_init(), maybe even init_timer() in case coreboot changes the way it implements timeouts in the future. Nothing is printed in this place, and I don't think die() would print anything, too.

krystian-hebel (changes requested):

src/soc/ibm/power9/romstage.c
SergiiDmytruk: Moved after `init_timer()` (amended last commit).

krystian-hebel (approved):

macpijan:
@krystian-hebel I think it's preferable to use `rebase` option when merging, so merge commit is not created. In fact, I have just disabled the merge commit option in repository settings.

But the best way (to keep the linter happy as well) is to use command line for merging, rather than github UI.

Something like:

```
git checkout raptor-cs_talos-2/develop
git merge --ff-only origin/setup_watchdog
git push origin raptor-cs_talos-2/develop
```

It will automatically close the MR in GitHub after pushing.

We might add it to dasharo docs I guess...

SergiiDmytruk:
@macpijan Dasharo/dasharo-issues#29 is still open, "fixes"-syntax doesn't seem to work across repositories. Should be closed now or before the next release?

PR #23, opened 2021-11-03
Simpler memlayout

krystian-hebel:


src/soc/ibm/power9/timer.c
```diff
@@ -58,6 +58,15 @@ void init_timer(void)
...
 
+	/*
+	 * 1. Data Cache Block set to Zero for 0..0x1000
+	 * 2. Instruction Cache Block Invalidate for 0..0x1000 - single 'isync'
+	 *    before enabling interrupts by writing to MSR is enough (not defined
+	 *    by ISA but in POWER9 Processor User's Manual, 4.6.2.2)
+	 */
+	for (tmp = 0; tmp < 0x1000; tmp += 128)
+		asm volatile("dcbz 0, %0; icbi 0, %0;" :: "r"(tmp) : "memory");
+
```
SergiiDmytruk, line 69: Could you mention that this initializes interrupt vectors and explain why this happens in `init_timer()`?

krystian-hebel: Done and rebased, however I may be able to simplify it more.

I've tried to make it work without interrupts and almost got it working. `wait` instruction waits until _exception_ occurs, without the need to actually take the _interrupt_ (on POWER exception is just a possible cause for interrupt, it doesn't always mean something bad as on x86). The only issue is that there is an exception at the start (maybe that asserted IPMI?) and I couldn't find a way to differentiate between them.

krystian-hebel: Ok, udelay() should work now and no longer needs separate implementation for ramstage

SergiiDmytruk (commented):
I didn't understand some details, but seems to make sense and to work fine.

krystian-hebel:
@SergiiDmytruk can you check if this doesn't break QEMU and mark this PR as approved?

SergiiDmytruk:
@krystian-hebel Sure, however some things on QEMU branch are being questioned (https://review.coreboot.org/c/coreboot/+/57082/12) and those comments seem applicable to this PR as well (extra stuff in ramstage). Please take a look at the review, maybe you'll want to apply suggested changes here as well or know why they are wrong. I'll try to follow reviewer's comments and see if everything still works in QEMU.

krystian-hebel:
If changes from here will work with QEMU with !RAMSTAGE half removed then it can be simplified too. All of those comments come from lack of understanding HRMOR, so we may need to add some documentation on that, but I'd start with testing before doing anything with comments from gerrit.

SergiiDmytruk:
Right, I was wondering why he didn't say anything about `BOOTBLOCK` or `ROMSTAGE` being always defined. I'll try explaining HRMOR in comments and see if it will satisfy Arthur.

SergiiDmytruk:
@krystian-hebel What modifications should suffice to make it work in QEMU? Changing `0xF8...` to `0x08` makes it start, but not for long:
```
coreboot-4.14-2461-gfc1e45d4ec-dirty Sat Nov 20 11:29:04 UTC 2021 bootblock starting (log level: 7)...
FMAP missing or corrupted at offset 0x20000!
Cannot locate primary CBFS
```

krystian-hebel:
I think this should be enough. What is your command line for starting QEMU? Does it have separate bootblock and CBFS?

SergiiDmytruk:
It's all in one ROM-file:
```
qemu-system-ppc64 -machine powernv,hb-mode=on \
                  -cpu power9 \
                  -m 4G \
                  -bios build/coreboot.rom \
                  -d unimp,guest_errors,mmu \
                  -D log.txt \
                  -serial stdio \
                  -kernel zImage.bundled \
                  -display none
```

krystian-hebel:
@SergiiDmytruk can you push your current state of QEMU platform to new branch? I'd like to debug it and it would be best if we have exactly the same code base.

SergiiDmytruk:
@krystian-hebel Pushed as `qemu-power9` with the last commit using changes from the PR.

krystian-hebel:
@SergiiDmytruk I tracked down the issue and added comment about how to fix it in https://review.coreboot.org/c/coreboot/+/57082. Other than that, in bootblock_crt0.S you should move code for saving FDT somewhere after HRMOR is modified and cache (re)initialized (but before %r3 is overwritten, or change all %r3 usages to another register) or Skiboot won't work. We may also use one of SPRs for saving this pointer (HSPRG?) - their content would be lost on hardware during deadman loop, but it doesn't have anything to pass in the first place.

SergiiDmytruk (approved):
Works fine in QEMU after fixing QEMU changes as suggested.

PR #25, opened 2021-11-05
Ramstage southbridge

miczyg1:


src/southbridge/amd/sb700/sm.c
```diff
@@ -0,0 +1,517 @@
...
+
+		byte |= 1 << 3;
+		pci_write_config8(dev, 0x43, byte);
+
+		/* Enable southbridge MMIO decode */
+		dword = pci_read_config32(dev, SB_MMIO_CFG_REG);
+		dword &= ~(0xffffff << 8);
+		dword |= SB_MMIO_BASE_ADDRESS;
+		dword |= 0x1;
+		pci_write_config32(dev, SB_MMIO_CFG_REG, dword);
```
krystian-hebel, line 305: This should be done in set_resources(). Right now set_resources() enables decoding without setting base address, which results in lack of serial output between those two points.

Note that I haven't checked if anything enabled by this register is actually used during that period and if it's safe to move it. Then again, accessing something under BAR before it is set wouldn't be good either.

PR #41, opened 2021-11-17
WIP: Northbridge fixes fixes

krystian-hebel:


src/device/resource_allocator_v4.c
```diff
@@ -198,15 +212,21 @@ static void compute_domain_resources(const struct device *domain)
 {
 	const struct device *child;
 	const int print_depth = 1;
+	struct bus *bus = domain->link_list;
 
 	if (domain->link_list == NULL)
 		return;
 
-	for (child = domain->link_list->children; child; child = child->sibling) {
+	while(bus && bus->children == NULL)
```
miczyg1, line 220: What will happen if bus is NULL instead of bus->children? `bus = bus->next;` will just access NULL pointer

src/device/resource_allocator_v4.c
```diff
@@ -428,6 +448,10 @@ static void avoid_fixed_resources(struct memranges *ranges, const struct device
 	}
 
 	bus = dev->link_list;
+
+	while(bus && bus->children == NULL)
+		bus = bus->next;
```
miczyg1, line 453: a/a

src/device/resource_allocator_v4.c
```diff
@@ -517,6 +541,9 @@ static void allocate_bridge_resources(const struct device *bridge)
 	struct device *child;
 	const unsigned long type_mask = IORESOURCE_TYPE_MASK | IORESOURCE_PREFETCH;
 
+	while(bus && bus->children == NULL)
+		bus = bus->next;
```
miczyg1, line 545: a/a

src/device/resource_allocator_v4.c
```diff
@@ -571,12 +598,15 @@ static void allocate_domain_resources(const struct device *domain)
 	struct memranges ranges;
 	struct device *child;
 	const struct resource *res;
+	struct bus *bus = domain->link_list;
+	while(bus && bus->children == NULL)
+		bus = bus->next;
```
miczyg1, line 603: a/a

src/device/resource_allocator_v4.c
```diff
@@ -658,11 +688,14 @@ static void allocate_domain_resources(const struct device *domain)
 void allocate_resources(const struct device *root)
 {
 	const struct device *child;
+	struct bus *bus = root->link_list;
+	while(bus && bus->children == NULL)
+		bus = bus->next;
```
miczyg1, line 693: a/a

miczyg1 (changes requested):

src/device/resource_allocator_v4.c
krystian-hebel: That is why there is test for `bus` before testing `bus->children`, neither the second test nor `bus->next` won't be accessed when `bus` is NULL. However, it does lack test for `bus` being NULL after this loop, where it blindly accesses `bus->children` in next `for` loop in most cases.

src/device/resource_allocator_v4.c
```diff
@@ -30,8 +30,16 @@ static const char *resource2str(const struct resource *res)
 
 static bool dev_has_children(const struct device *dev)
 {
-	const struct bus *bus = dev->link_list;
-	return bus && bus->children;
+	const struct bus *link = dev->link_list;
+
+	if (!link)
```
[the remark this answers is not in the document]
krystian-hebel: Yes, it will, this is leftover from more verbose version, I'll remove it.

PR #43, opened 2021-11-22
Vboot enablement

miczyg1:


macpijan:
@miczyg1 maybe you need rebase, because the build is failing (not latest CI configuration I think)

PR #60, opened 2021-12-15
.github/workflows/build.yml: clean up working tree before building KG…

miczyg1:
…PE-D16


miczyg1:
Now the built binaries are not marked dirty

PR #62, opened 2021-12-20
WIP: cbmem: Fix endianness on BE systems

IgorBagnucki:
Signed-off-by: Igor Bagnucki <igor.bagnucki@3mdeb.com>

### DIFF
--- a/util/cbmem/cbmem.c
+++ b/util/cbmem/cbmem.c
@@ -230,17 +230,17 @@ static int find_cbmem_entry(uint32_t id, uint64_t *addr, size_t *size)
 		const struct lb_cbmem_entry *lbe;
 
 		lbr = (const void *)(table + offset);
-		offset += lbr->size;
+		offset += FIX_ENDIANNESS(lbr->size);
 
-		if (lbr->tag != LB_TAG_CBMEM_ENTRY)
+		if (FIX_ENDIANNESS(lbr->tag) != LB_TAG_CBMEM_ENTRY)
 			continue;
 
 		lbe = (const void *)lbr;
-		if (lbe->id != id)
+		if (FIX_ENDIANNESS(lbe->id) != id)
 			continue;
 
-		*addr = lbe->address;
-		*size = lbe->entry_size;
+		*addr = FIX_ENDIANNESS(lbe->address);
+		*size = FIX_ENDIANNESS(lbe->entry_size);
 		ret = 0;
 		break;
 	}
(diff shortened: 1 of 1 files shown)


util/cbmem/cbmem.c
```diff
@@ -22,12 +22,18 @@
...
 
+#include <byteswap.h>
+
 #ifdef __OpenBSD__
 #include <sys/param.h>
 #include <sys/sysctl.h>
 #endif
 
 #define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))
+#define FIX_ENDIANNESS(a) (sizeof(a) == 1 ? (a) :\
```
krystian-hebel, line 33: This is definitely not a proper way of doing it. We cannot assume that FW endianness will always be different than OS, such approach will break on all x86. Even on PPC64 we could possibly have BE OS, so we have to know when to swap bytes and when not. This needs some redesign of coreboot tables because as of now there is no indication about endianness saved in them.

There are some functions that can convert from BE to host, like `ntohl()` used in `dt_update_cells()`, they can be used instead of `bswap`.

util/cbmem/cbmem.c
```diff
@@ -116,8 +122,8 @@ static const void *map_memory(struct mapping *mapping, unsigned long long phys,
 			phys - mapping->offset);
 
 	if (v == MAP_FAILED) {
-		debug("Mapping failed %zuB of physical memory at 0x%llx.\n",
-			mapping->virt_size, phys - mapping->offset);
+		debug("Mapping failed %zuB of physical memory at 0x%llx.\n Error: %s\n",
+			mapping->virt_size, phys - mapping->offset, strerror(errno));
```
krystian-hebel, line 126: Have you checked if `strerror(errno)` is available on BSD systems?

krystian-hebel (changes requested):

util/cbmem/cbmem.c
IgorBagnucki: Agree, this cannot be finally implemented like this. I have created a macro to make it easier to fix later making necessary checks in just one place instead of many.

util/cbmem/cbmem.c
IgorBagnucki: Should be available as it is documented.
* [strerror doc](https://www.freebsd.org/cgi/man.cgi?query=strerror&apropos=0&sektion=0&manpath=FreeBSD+6.2-RELEASE&format=html)
* [errno doc](https://www.freebsd.org/cgi/man.cgi?query=errno&sektion=2&manpath=freebsd-release-ports

Additionally this function is already used in cbmem [here](https://github.com/Dasharo/coreboot/blob/cc39cf71072a4da2afa828d30e0031fba46f95be/util/cbmem/cbmem.c#L450).

krystian-hebel: Right, but at that point it is protected with `#if defined(linux)`. We'll have to see if CI likes it or not.

util/cbmem/cbmem.c
SergiiDmytruk: > This needs some redesign of coreboot tables because as of now there is no indication about endianness saved in them.

Looking at `sys/arch` Power seems to be the first non-LE architecture, so one way of doing it could be to always use LE for tables. Another way is to use a heuristic like PR already does: `header_bytes` and `header_checksum` in wrong endianness are easily detectable. In both cases compatibility is preserved.

"LBIO" could also be reversed if structure uses BE. And if it will always LE so far, size can be made 1 or 2 bytes and the rest would become flags to indicate endianness and something else in the future.

macpijan:
@SergiiDmytruk Please confirm, but AFAIK the relevan part of this is already on https://github.com/Dasharo/coreboot/commits/raptor-cs_talos-2/release and this one can be closed.

SergiiDmytruk:
Correct, https://github.com/Dasharo/coreboot/pull/197 replaced this PR.

PR #56, opened 2021-12-12
Implement FSI for Power9

SergiiDmytruk:
Not much tested, but this works and prints zeros:
```c
printk(BIOS_EMERG, "read_fsi(0, 0x3418) = 0x%08x\n", read_fsi(0, 0x003400 | 0x18));
```
and this hangs, which is probably expected if there are actually no slaves on MFSI of master CPU:
```c
printk(BIOS_EMERG, "read_fsi(1, 0x3418) = 0x%08x\n", read_fsi(1, 0x003400 | 0x18));
```

PR #64, opened 2021-12-24
Implement isteps 8.1-8.4

SergiiDmytruk:
This implements initialization of second CPU in the system and starts its SBE.

Depends on #57 and transitively on #56 PRs.


krystian-hebel:
I've changed base branch so it will be easier to review, I'll do the same with other PRs. I think I will merge it from the newest to oldest to keep history clean.

SergiiDmytruk:
> I think I will merge it from the newest to oldest to keep history clean.

I could just rebase, but either way will work.

src/soc/ibm/power9/istep_8_1.c
```diff
@@ -0,0 +1,337 @@
...
+	 *       see getObusPllBucket() in Hostboot for values of ATTR_OB*_PLL_BUCKET */
+
+	write_cfam(chip, MBOX_SCRATCH_REG1_FSI + 1, scratch >> 32);
+
+	/* SCRATCH_REGISTER_3 */
+
+	scratch = PPC_SHIFT(read_cfam(chip, MBOX_SCRATCH_REG1_FSI + 2), 31);
+
+	boot_flags = (read_scom(MBOX_SCRATCH_REG1 + 2) >> 32);
+	risk_level = (get_dd() < 0x23 ? 0 : 4);
```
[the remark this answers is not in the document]
SergiiDmytruk: > maybe that is the reason for stuck CPU frequency

Lack of sensor information looks like a more plausible reason to me. OCC likely needs them to determine CPU load.

> For some reason this is 0 in scratch register for first CPU, even though it is DD23. 

I read somewhere that first CPU is started by BMC, maybe it doesn't know proper risk level. However since Hostboot contains code for upgrading it
```
2339.91027|HWAS_I|I>hwas.C: updateProcCompatibilityRiskLevel: Update RISK_LEVEL from 0 to 4 (EC Level: 0x23, RISK_ORIGIN: MRW, PROC_COMPATIBILITY_REQ: 0)
```
it's probably expected (there is also `NIMBUS_ONLY` in comments).

> Also, this line sets risk level for second CPU based on DD of first CPU, and they may not be the same.

Risk level comes from a *system* attribute, so I guess CPUs should have identical level.

> If it is even possible at this point, only FSI SCOM can be used to read from second CPU.

Haven't tried with that address, but I think FSI SCOM needs CPU to be on. Easy to check though.

src/soc/ibm/power9/istep_8_1.c
```diff
@@ -0,0 +1,337 @@
...
+#include <arch/io.h>
+#include <stdint.h>
+#include <string.h>
+#include <console/console.h>
+#include <cpu/power/istep_8.h>
+#include <cpu/power/mvpd.h>
+#include <cpu/power/scom.h>
+
+#include "fsi.h"
+#include "homer.h"
```
krystian-hebel, line 12: This is only for `IS_*_FUNCTIONAL`? If this is the case then it should probably be put into another, smaller header. Probably together with `get_available_cores()`, and this function should be used in homer.c (this may be done later, when working on secondary HOMER & OCC stuff). The way I did it seems too hacky, and may not work for second CPU anyway.

src/soc/ibm/power9/istep_8_1.c
```diff
@@ -0,0 +1,337 @@
...
+	enum {
+		VPD_CP00_PG_HDR_LENGTH   = 1,
+		VPD_CP00_PG_DATA_LENGTH  = 128,
+		VPD_CP00_PG_DATA_ENTRIES = VPD_CP00_PG_DATA_LENGTH / 2,
+
+		ALL_ON_PG_MASK = 0xFFFF,
+		EC_AG_MASK     = 0xE1FF,
+	};
+
+	uint64_t cores = 0;
```
krystian-hebel, line 37: If this function is made reusable, this should be made static table to avoid reading MVPD multiple times. Doesn't have to be in this PR, though.

src/soc/ibm/power9/istep_8_1.c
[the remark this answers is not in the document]
krystian-hebel: Ok, so it seems to be right thing to do. I wonder if we should update scratch register for CPU0 or if it is ok to leave it as it is set by BMC.

src/soc/ibm/power9/istep_8_1.c
SergiiDmytruk: > This is only for `IS_*_FUNCTIONAL`?

Yes. Macros for checks and `MAX_*_PER_CHIP` constants are two common reasons for including `homer.h`.

src/soc/ibm/power9/istep_8_1.c
SergiiDmytruk: In changes related to HOMER I moved it to `mvpd.c` and made public (it's still used only for one CPU and called only once in romstage and once in ramstage).

krystian-hebel (approved):

PR #71, opened 2022-01-11
project_config.yaml: Fix url template for skiboot

IgorBagnucki:
Signed-off-by: Igor Bagnucki <igor.bagnucki@3mdeb.com>


macpijan:
We do not use `coreboot` repository for storing these YAMLs abymore

PR #77, opened 2022-01-18
Initialize PCIe for the second CPU

SergiiDmytruk:
Updates isteps 10.10, 10.12 and 14.3 with CPU-specific configuration and chip-specific SCOM accesses.


src/soc/ibm/power9/istep_10_10.c
```diff
@@ -161,14 +161,37 @@ static const size_t pec_lane_cfg_sizes[] = {
...
+static uint16_t lane_masks_chip1[MAX_PEC_PER_PROC][MAX_LANE_GROUPS_PER_PEC] = {
+	{ LANE_MASK_X16,     0x0,               0x0,               0x0 },
+	{ LANE_MASK_X8_GRP0, 0x0, LANE_MASK_X8_GRP1,               0x0 },
+	{ LANE_MASK_X16,     0x0,               0x0,               0x0 },
+};
+
+static uint16_t (*lane_masks[])[MAX_LANE_GROUPS_PER_PEC]  = {
+	lane_masks_chip0,
+	lane_masks_chip1,
+};
```
krystian-hebel, line 182: This is hard to read and gives unnecessary level of indirection.
```suggestion
static uint16_t lane_masks[][MAX_PEC_PER_PROC][MAX_LANE_GROUPS_PER_PEC]  = {
	{
		{ LANE_MASK_X16,     0x0,               0x0,               0x0 },
		{ LANE_MASK_X8_GRP0, 0x0, LANE_MASK_X8_GRP1,               0x0 },
		{ LANE_MASK_X8_GRP0, 0x0, LANE_MASK_X4_GRP0, LANE_MASK_X4_GRP1 },
	},
	{
		{ LANE_MASK_X16,     0x0,               0x0,               0x0 },
		{ LANE_MASK_X8_GRP0, 0x0, LANE_MASK_X8_GRP1,               0x0 },
		{ LANE_MASK_X16,     0x0,               0x0,               0x0 },
	},
};
```

src/soc/ibm/power9/istep_10_10.c
```diff
@@ -161,14 +161,37 @@ static const size_t pec_lane_cfg_sizes[] = {
...
+
+/*
+ * PROC_PCIE_IOP_SWAP from processed talos.xml for each PEC of each processor
+ * chip
+ */
+static uint8_t pcie_iop_swap_chip0[MAX_PEC_PER_PROC] = { 1, 0, 0 };
+static uint8_t pcie_iop_swap_chip1[MAX_PEC_PER_PROC] = { 1, 0, 4 };
+
+static uint8_t *pcie_iop_swap[] = {
```
krystian-hebel, line 191: Same as above.

src/soc/ibm/power9/istep_10_10.c
```diff
@@ -574,24 +606,25 @@ static void phase1(const struct lane_config_row **pec_cfgs,
...
-void istep_10_10(uint8_t *phb_active_mask, uint8_t *iovalid_enable)
+void istep_10_10(uint8_t chips, struct pci_info *pci_info)
 {
-	const struct lane_config_row *pec_cfgs[MAX_PEC_PER_PROC] = { NULL };
-
 	printk(BIOS_EMERG, "starting istep 10.10\n");
 	report_istep(10,10);
 
-	determine_lane_configs(phb_active_mask, pec_cfgs);
+	for (uint8_t chip = 0; chip < 8; chip++) {
```
krystian-hebel, line 614:
```suggestion
	for (uint8_t chip = 0; chip < 2; chip++) {
```

`pci_info` only has 2 elements so we shouldn't make the code able to read out of bounds, coverity scan could complain about it.

src/soc/ibm/power9/istep_10_12.c
```diff
@@ -7,8 +7,8 @@
 
 #include "pci.h"
 
-/* PCIe only at the moment, should also do other buses */
-static void enable_ridi(void)
+/* PCIe only at the moment, Hostboot also updatse MC and OBus chiplets too */
```
krystian-hebel, line 10:
```suggestion
/* PCIe only at the moment, Hostboot also updates MC and OBus chiplets too */
```

src/soc/ibm/power9/istep_10_12.c
```diff
@@ -21,23 +21,26 @@ static void enable_ridi(void)
...
 }
 
-void istep_10_12(void)
+void istep_10_12(uint8_t chips)
 {
 	printk(BIOS_EMERG, "starting istep 10.12\n");
 	report_istep(10,12);
 
-	enable_ridi();
+	for (uint8_t chip = 0; chip < 8; chip++) {
```
krystian-hebel, line 40:
```suggestion
	for (uint8_t chip = 0; chip < 2; chip++) {
```

src/soc/ibm/power9/istep_14_3.c
```diff
@@ -473,29 +476,34 @@ static void init_phbs(uint8_t phb_active_mask, const uint8_t *iovalid_enable)
...
 
-void istep_14_3(uint8_t phb_active_mask, const uint8_t *iovalid_enable)
+void istep_14_3(uint8_t chips, const struct pci_info *pci_info)
 {
 	printk(BIOS_EMERG, "starting istep 14.3\n");
 	report_istep(14,3);
 
-	init_pecs(iovalid_enable);
-	init_phbs(phb_active_mask, iovalid_enable);
+	for (uint8_t chip = 0; chip < 8; chip++) {
```
krystian-hebel, line 500:
```suggestion
	for (uint8_t chip = 0; chip < 2; chip++) {
```

src/soc/ibm/power9/istep_10_10.c
krystian-hebel: Or even better, `#define` this value and use it where appropriate, in case there will ever be a board with more than 2 CPU sockets.

SergiiDmytruk:
In addition to addressing comments also made 14.3 use chip for group ID in one place (the other one is address extension  group ID which is always zero). I had this on OCC branch, but didn't apply here until now.

src/soc/ibm/power9/istep_14_3.c
```diff
@@ -277,26 +281,26 @@ static void init_phbs(uint8_t phb_active_mask, const uint8_t *iovalid_enable)
...
 	/* Determine base address of chip MMIO range */
 	uint64_t base_addr_mmio = 0;
-	base_addr_mmio |= PPC_PLACE(0, 8, 5);  // ATTR_PROC_FABRIC_SYSTEM_ID
-	base_addr_mmio |= PPC_PLACE(0, 15, 4); // ATTR_PROC_EFF_FABRIC_GROUP_ID
-	base_addr_mmio |= PPC_PLACE(0, 19, 3); // ATTR_PROC_EFF_FABRIC_CHIP_ID
-	base_addr_mmio |= PPC_PLACE(3, 13, 2); // FABRIC_ADDR_MSEL
-	                                       // nm = 0b00/01, m = 0b10, mmio = 0b11
+	base_addr_mmio |= PPC_PLACE(0, 8, 5);     // ATTR_PROC_FABRIC_SYSTEM_ID
+	base_addr_mmio |= PPC_PLACE(chip, 15, 4); // ATTR_PROC_EFF_FABRIC_GROUP_ID
+	base_addr_mmio |= PPC_PLACE(0, 19, 3);    // ATTR_PROC_EFF_FABRIC_CHIP_ID
```
krystian-hebel, line 296: Is it always set this way or does it depend on `MBOX_SCRATCH_REG6_GROUP_PUMP_MODE`?

In any case please add a comment mentioning that chip is not written to CHIP_ID and this is not an error (unless it is).

SergiiDmytruk: Not an error. Depends on group pump mode.

I was wondering whether pump mode can vary for Nimbus at all and thus whether it makes sense to support both options? If both are potentially relevant, might as well extract a function like for DD, because the check is needed in multiple places and their number grows. If both are necessary, we'll need to verify that the check wasn't skipped anywhere else (compare against HB uses of the corresponding attribute).

krystian-hebel (approved):

PR #80, opened 2022-01-31
mb/clevo/tgl-u/variants/nv41mz/devicetree.cb: set HDA subsystem ID

mkopec:
Set correct subsystem ID for HD Audio device.
Needed for Windows drivers to load correct jack configuration.
Fixes [microphone input on Windows.](https://gitlab.com/novacustom/dasharo-compatibility/-/issues/15#note_822393261)

### DIFF
--- a/src/drivers/gfx/nvidia/optimus/optimus.c
+++ b/src/drivers/gfx/nvidia/optimus/optimus.c
@@ -60,16 +60,33 @@ static void nvidia_optimus_acpi_l23_entry(void)
 }
 
 /* Restore the subsystem ID after _ON to ensure the windows driver works */
-static void nvidia_optimus_acpi_subsystem_id_restore(void)
+static void nvidia_optimus_acpi_subsystem_id_restore(const struct device *dev)
 {
-	uint32_t dvid = CONFIG_SUBSYSTEM_DEVICE_ID << 16 | CONFIG_SUBSYSTEM_VENDOR_ID;
+	uint32_t did, vid, dvid;
+
+	if (CONFIG_SUBSYSTEM_VENDOR_ID)
+		vid = CONFIG_SUBSYSTEM_VENDOR_ID;
+	else if (dev->subsystem_vendor)
+		vid = dev->subsystem_vendor;
+	else
+		vid = 0x0000;
+
+	if (CONFIG_SUBSYSTEM_DEVICE_ID)
+		did = CONFIG_SUBSYSTEM_DEVICE_ID;
+	else if (dev->subsystem_device)
+		did = dev->subsystem_device;
+	else
+		did = 0x0000;
+
+	dvid = did << 16 | vid;
 	acpigen_write_store_int_to_namestr(dvid, ACPI_REG_PCI_SUBSYSTEM_ID);
 }
 
 static void
 nvidia_optimus_acpi_method_on(unsigned int pcie_rp,
-			 const struct drivers_gfx_nvidia_optimus_config *config)
+			      const struct device *dev)
 {
+	const struct drivers_gfx_nvidia_optimus_config *config = config_of(dev);
 	acpigen_write_method_serialized("_ON", 0);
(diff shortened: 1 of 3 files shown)


miczyg1 (approved):

miczyg1:
Looks ok, however I am not sure how it will react to the Kconfig SUBSYSTEM_ID optionset

I would add `subsystemid 0x1558 0x4018 inherit` in the PCI domain scope and override it per device in the devicetree how yo did it for HDA. Would have to check static.c and final system lspci logs to check whether the subsystem is good.

mkopec:
> I suspect we have subsystem hardcoded in Nvidia dGPU. We would have to make it dynamic with SSDT...

Actually we're already setting it in SSDT: https://github.com/Dasharo/coreboot/blob/538c29bcada9b8b9d2bad2c4b45e53dbcd1f941d/src/drivers/gfx/nvidia/optimus/optimus.c#L62

though we're using the `CONFIG_SUBSYSTEM_*` options there, would it still work if I set it in devicetree.cb instead of Kconfig?

miczyg1:
There are two options:
1. Check static.c whether dGPU device populate `subsystem_vendor` and `subsystem_device` in the `struct device` of the dGPU. You could use them instead of Kconfig.
2. Add `subsystem_vendor` and `subsystem_device` to the `drivers_gfx_nvidia_optimus_config` and use it.

tl;dr
Shouldn't we check first whether subsystem is set in the Kconfig (or devicetree) before we emit them here https://github.com/Dasharo/coreboot/blob/538c29bcada9b8b9d2bad2c4b45e53dbcd1f941d/src/drivers/gfx/nvidia/optimus/optimus.c#L66 ?

mkopec:
>     1. Check static.c whether dGPU device populate `subsystem_vendor` and `subsystem_device` in the `struct device` of the dGPU. You could use them instead of Kconfig. PREFERRED

I took this approach. Please review

> 
> tl;dr Shouldn't we check first whether subsystem is set in the Kconfig (or devicetree) before we emit them here
> 
> https://github.com/Dasharo/coreboot/blob/538c29bcada9b8b9d2bad2c4b45e53dbcd1f941d/src/drivers/gfx/nvidia/optimus/optimus.c#L66
> ?

We should. Added a check as well.

miczyg1:
@mkopec looks good now. And the most important question: does it work? :)

mkopec:
@miczyg1 Yes, it still works :) the nvidia driver is still working on windows and the correct subsystem ID is set in device manager. The SD card reader still has the wrong subsystem ID though.

miczyg1:
```
		device ref pcie_rp9 on
			# PCIe root port #9 x1, Clock 3 (CARD)
			register "PcieRpEnable[8]" = "1"
			register "PcieRpLtrEnable[8]" = "1"
			register "PcieClkSrcUsage[3]" = "8"
			register "PcieClkSrcClkReq[3]" = "3"
			device pci 0.0 on end # SD card reader, to populate subsystem
		end
```

@mkopec check if something like this would do the work

mkopec:
This seems to work according to coreboot logs: 

```
PCI: 2e:00.0 subsystem <- 1558/4018
```

however, the subsystem id in lspci is still the default value: 

```
2e:00.0 Unassigned class [ff00]: Realtek Semiconductor Co., Ltd. RTS522A PCI Express Card Reader [10ec:522a] (rev 01)
	Subsystem: Realtek Semiconductor Co., Ltd. RTS522A PCI Express Card Reader [10ec:522a]
	Flags: bus master, fast devsel, latency 0, IRQ 145
	Memory at 80500000 (32-bit, non-prefetchable) [size=4K]
	Capabilities: <access denied>
	Kernel driver in use: rtsx_pci
	Kernel modules: rtsx_pci
```

same result on Windows too.

miczyg1:
Hmmm... I don't know why, but it can be blocked. I wonder if we can try to set it early before FSP-S...

miczyg1:
Anyway this is subject to other PR. Merging this one.

PR #82, opened 2022-02-02
soc/power9/mvpd.c: cache MVPD in memory between stages

krystian-hebel:
MVPD is now read only once per CPU. I haven't checked how much time this saves exactly due to big fluctuations of serial output, but it is noticeably faster.


SergiiDmytruk (approved):
ramstage is faster now, thanks! I've removed now outdated comment and renamed `MVPD_SIZE` to `MAX_MVPD_SIZE` before merging.

PR #83, opened 2022-02-03
Clevo ns50mu/create dirs

mkopec:
Add Clevo NS50MU as a variant of the Clevo TGL-U baseboard.

For now the mainboard is just a copy of the `nv41mz` variant.


src/mainboard/clevo/tgl-u/bootblock.c
```diff
@@ -11,6 +11,7 @@
 #define DGPU_RST_N GPP_U4
 #define DGPU_PWR_EN GPP_U5
 
+#ifdef DRIVER_NVIDIA_OPTIMUS
 static void dgpu_power_enable(int onoff) {
```
miczyg1, line 15: This function could be exposed by a header in the nvidia driver. Then I wouldn't need to be guarded by preprocessor

src/mainboard/clevo/tgl-u/bootblock.c
```diff
@@ -34,5 +36,7 @@ void bootblock_mainboard_early_init(void)
 
 	pads = variant_early_gpio_table(&num);
 	gpio_configure_pads(pads, num);
+#ifdef DRIVER_NVIDIA_OPTIMUS
 	dgpu_power_enable(1);
```
miczyg1, line 40: If this function is exposed by a header in nvidia driver, then
`if (CONFIG(DRIVER_NVIDIA_OPTIMUS))` would be enough

src/mainboard/clevo/tgl-u/variants/ns50mu/devicetree.cb
```diff
@@ -0,0 +1,365 @@
+chip soc/intel/tigerlake
```
miczyg1, line 1: Can we make a basebaord devicetree and add overrides here since we have more than one variant? overrides should be significantly smaller and avoid duplication of the code

src/mainboard/clevo/tgl-u/variants/ns50mu/gpio.c
```diff
@@ -0,0 +1,259 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
```
miczyg1, line 1: Any differences between nv41?

src/mainboard/clevo/tgl-u/variants/ns50mu/hda_verb.c
```diff
@@ -0,0 +1,41 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
```
miczyg1, line 1: any differences between nv41 and this board?

src/mainboard/clevo/tgl-u/variants/ns50mu/romstage.c
```diff
@@ -0,0 +1,39 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
```
miczyg1, line 1: Does it need a separate file?

miczyg1 (changes requested):

src/mainboard/clevo/tgl-u/variants/ns50mu/gpio.c
mkopec: Yes, please see https://github.com/Dasharo/coreboot/pull/84

src/mainboard/clevo/tgl-u/variants/ns50mu/devicetree.cb
mkopec: Yes, we can do this. I was planning to first develop a devicetree for NS50MU based on the NV41MZ devicetree and then find the common parts and move them to a baseboard devicetree.

I wanted to create a basic structure in this PR and then do the actual port in https://github.com/Dasharo/coreboot/pull/84

src/mainboard/clevo/tgl-u/variants/ns50mu/romstage.c
mkopec: Probably not, I'll move it to baseboard code

src/mainboard/clevo/tgl-u/variants/ns50mu/hda_verb.c
mkopec: Yes, the subsystem ID and one line of pin config are different

src/mainboard/clevo/tgl-u/variants/ns50mu/romstage.c
mkopec: Moved

src/mainboard/clevo/tgl-u/bootblock.c
mkopec: i used `if (CONFIG(DRIVER_NVIDIA_OPTIMUS))` instead and removed all the `#ifdefs`

src/mainboard/clevo/tgl-u/bootblock.c
mkopec: Removed the `#ifdef` since it's not actually needed

miczyg1: I meant define `void dgpu_power_enable(int onoff) ` in a header and let the mainboard implement it. without static you may implement it but not necessarily use

PR #85, opened 2022-02-04
hide different SCOM implementations behind common API

krystian-hebel:


krystian-hebel:
@SergiiDmytruk I've reviewed and tested this after your changes, but GH won't allow me to click approve button, can you do it and then merge?

SergiiDmytruk (approved):

PR #86, opened 2022-02-07
ec/clevo/it5570: color keyboard backlight

mkopec:


mkopec:
Looks like the backlight state isn't restored on wake, we're probably doing it too early.

miczyg1:
Maybe putting it to https://github.com/Dasharo/coreboot/blob/504840bcd861d149bcfb66b6dafa7b4b0db60b58/src/mainboard/clevo/tgl-u/acpi/sleep.asl#L37
would be better?

mkopec:
@miczyg1 Moving it to MS0X didn't help. Adding a `Sleep(1000)` before KBUP () doesn't help either.

mkopec:
Superseded by https://github.com/Dasharo/coreboot/pull/136

PR #90, opened 2022-02-09
Enable RAM local to the second CPU

SergiiDmytruk:
This updates both MCs and DRAM training isteps and whatever else was necessary to make all memory available. The code base didn't lend itself to splitting this in two PRs because of common functions. Debian started with these changes sees all installed RAM.

Maybe pay attention to https://github.com/Dasharo/coreboot/commit/4f43da532fb3c550d6b9365d754e46da41d68c34 in review, I've made the changes thinking that something is wrong there while another part was causing issues, but everything worked without these changes as well.


krystian-hebel:
Of course those `dp` functions were wrong, nice catch.

src/soc/ibm/power9/mvpd.c
```diff
@@ -65,7 +67,7 @@ struct pt_record {
  * of bytes read. */
 static int read_eeprom_chip(uint8_t cpu, uint32_t offset, void *data, uint16_t len)
 {
-	const unsigned int bus = 1 + 4 * cpu; // four I2C buses per CPU
+	const unsigned int bus = (cpu == 0 ? 1 : FSI_I2C_BUS);
```
krystian-hebel, line 70: Making sure: this will always be FSI for second CPU because MVPD is needed before XSCOM is initialized and all later code uses MVPD from cache?

src/soc/ibm/power9/i2c.c
```diff
@@ -93,65 +98,72 @@ void get_spd_smbus(struct spd_block *blk)
 	blk->len = SPD_PAGE_LEN_DDR4;
 }
 
+void get_spd_smbus(struct spd_block *blk)
+{
+	get_spd_smbus_ext(/*chip=*/0, blk);
+}
+
```
krystian-hebel, line 105: Not needed.

krystian-hebel (changes requested):
Some of suggested changes are to remove obsolete TODOs. One delay was unintentionally (I hope) removed and should be restored.

src/include/cpu/power/mvpd.h
```diff
@@ -57,6 +57,10 @@ const struct region_device *mvpd_device_ro(void);
 /* Reads #V of one of LRP records (mind that there is only one buffer) */
 const struct voltage_kwd *mvpd_get_voltage_data(uint8_t chip, int lrp);
 
+/* Sets pg[0] and pg[1] to partial good values for MC01_CHIPLET_ID and
+ * MC23_CHIPLET_ID respectively */
```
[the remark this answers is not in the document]
SergiiDmytruk: It's read from MVPD which is built out of EEPROM data, so I guess it's set on a factory.

src/soc/ibm/power9/chip.c
```diff
@@ -487,28 +488,38 @@ static inline unsigned long size_k(uint64_t reg)
...
+		for (mcs_i = 0; mcs_i < MCS_PER_PROC; mcs_i++) {
+			uint64_t reg;
+			chiplet_id_t nest = mcs_to_nest[mcs_ids[mcs_i]];
+
+			/* These registers are undocumented, see istep 14.5. */
+			/* MCS_MCFGP */
+			reg = read_rscom_for_chiplet(chip, nest, 0x0501080A);
+			if (reg & PPC_BIT(0)) {
+				uint64_t end = base_k(reg) + size_k(reg);
+				ram_resource(dev, idx++, base_k(reg), size_k(reg));
```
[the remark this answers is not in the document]
SergiiDmytruk: Base address includes group (chip) ID (or'ed with `0x0000200000000000` for the second CPU).

src/soc/ibm/power9/i2c.c
```diff
@@ -39,42 +40,46 @@
...
 
 #define I2C_MAX_FIFO_CAPACITY	8
-#define SPD_I2C_BUS		3
 
 enum i2c_type {
-	HOST_I2C, // I2C via XSCOM (first CPU)
-	FSI_I2C,  // I2C via FSI (second CPU)
+	HOST_I2C_CPU0, // I2C via XSCOM (first CPU)
+	HOST_I2C_CPU1, // I2C via XSCOM (second CPU)
+	FSI_I2C,       // I2C via FSI (second CPU)
```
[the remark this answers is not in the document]
SergiiDmytruk: It's similar: FSI version is always available and host version becomes usable along with XSCOM (SBE SCOM would do to). FSI I2C has only one bus/engine, not 4 as host I2C, but we need only one to read MVPD initially.

src/soc/ibm/power9/mvpd.c
SergiiDmytruk: Yes, except that it works without caching too (probably slower than it would with XSCOM).

src/soc/ibm/power9/istep_14_5.c
```diff
@@ -257,81 +294,63 @@ static void fir_unmask(int mcs_i)
...
+
+void istep_14_5(uint8_t chips)
+{
+	uint8_t chip;
+
+	printk(BIOS_EMERG, "starting istep 14.5\n");
+	report_istep(14,5);
+
+	/* Start MCS reset */
+	revert_mc_hb_dcbz_config(/*chip=*/0);
```
krystian-hebel, line 348: This is called in `proc_setup_bars()`, why also here?

src/soc/ibm/power9/romstage.c
```diff
@@ -76,31 +76,32 @@ static inline bool is_proper_dimm(spd_raw_data spd, int slot)
...
+	mem_data[chip].mcs[mcs].mca[mca].functional = false;
 
 	/* Propagate upwards */
-	if (mem_data.mcs[mcs].mca[mca ^ 1].functional == false) {
-		mem_data.mcs[mcs].functional = false;
-		if (mem_data.mcs[mcs ^ 1].functional == false)
+	if (mem_data[chip].mcs[mcs].mca[mca ^ 1].functional == false) {
+		mem_data[chip].mcs[mcs].functional = false;
+		if (mem_data[chip].mcs[mcs ^ 1].functional == false)
 			die("No functional MCS left");
```
krystian-hebel, line 87: This is not critical error with 2 CPUs, one of them can potentially have no DIMMs installed
```suggestion
			printk(BIOS_INFO, "No functional MCS left on chip %d", chip);
```

Alternatively we can check it one level higher, but I haven't analyzed if this can break 1 CPU systems. It is also tested in `prepare_dimm_data()` and it should `die()` there anyway so ordinary print should be enough.

src/soc/ibm/power9/istep_14_5.c
SergiiDmytruk: I meant to move it from `proc_setup_bars()`, HB calls it only for the master CPU.

krystian-hebel: Makes sense, other CPUs shouldn't have their cache initialized by SBE.

src/soc/ibm/power9/romstage.c
SergiiDmytruk: Made it an info message and also updated `prepare_dimm_data()` to use loops.

krystian-hebel (approved):
Just one small issue left. I will merge it and add one commit to fix it so we won't have to go through another round of review.

PR #92, opened 2022-02-11
soc/intel/tigerlake: hide IOM and PMC devices in ACPI

mkopec:
Hide devices that appear as `Unknown Devices` in Windows Device Manager.


miczyg1 (approved):
Do we observe the same behaviour on the original firmware?

mkopec:
@miczyg1 Yes, both devices aren't visible on stock firmware.

PR #93, opened 2022-02-12
Build HOMER image for the second CPU

SergiiDmytruk:
* Re-organize code for building HOMER a bit in preparation for building multiple images.
* Update SCOM accesses.
* Update various ways of accessing or producing data derived from MVPD.


src/soc/ibm/power9/homer.c
```diff
@@ -3294,43 +3266,76 @@ uint64_t build_homer_image(void *homer_bar)
...
-	/* 15.3 establish EX chiplet */
-	report_istep(15,3);
-	/* Multicast groups for cores were assigned in get_available_cores() */
-	for (int i = 0; i < MAX_QUADS_PER_CHIP; i++) {
-		if (IS_EQ_FUNCTIONAL(i, cores) &&
-		    (read_scom_for_chiplet(EP00_CHIPLET_ID + i, 0xF0001) & PPC_BITMASK(3,5))
-		    == PPC_BITMASK(3,5))
-			scom_and_or_for_chiplet(EP00_CHIPLET_ID + i, 0xF0001,
-			                        ~(PPC_BITMASK(3,5) | PPC_BITMASK(16,23)),
+	write_rscom(chip, 0x05012B02, (uint64_t)homer + (8 - chip) * 4 * MiB);		// FIXME
```
krystian-hebel, line 3280: This address (OCC common area or whatever its official name is) should be passed by the caller.

I think we should rework calculation for HOMER base. I hardcoded it to the way Hostboot sets it so I can reuse its image but that is no longer necessary. We can reserve only as much as is needed (8M + 4M * number of chips), instead of always reserving 40M (IIRC). This can be done now or as separate change later, whichever is easier.

krystian-hebel (commented):
There are `TODO: don't hard-code chip number` comments that are addressed in further PRs (#94, #101) so it may be better to merge mentioned PRs into this one first.

src/soc/ibm/power9/homer.c
SergiiDmytruk: I don't think it will complicate rebasing, so done it here.

SergiiDmytruk:
> There are `TODO: don't hard-code chip number` comments that are addressed in further PRs (#94, #101) so it may be better to merge mentioned PRs into this one first.

Those are just reminders after changing function prototypes, code still works as before. And merging in different order won't get rid of them in commit history anyway.

I've added a commit to fix I2C, your change allowed only bus number 0 not just for FSI I2C, but for I2C in general.

krystian-hebel:
You are absolutely right about I2C, I must have boot-tested wrong image.

I suggested merging other PRs first because I tried to have `develop` in more or less useful state, but since I messed up anyway it may be merged. I will do one more boot test before final approve, hopefully correctly this time.

src/soc/ibm/power9/istep_13_8.c
```diff
@@ -33,7 +33,8 @@
  */
 static void p9n_mca_scom(uint8_t chip, int mcs_i, int mca_i)
 {
-	const struct powerbus_cfg *pb_cfg = powerbus_cfg();
+	/* FIXME: don't hard-code chip number. */
+	const struct powerbus_cfg *pb_cfg = powerbus_cfg(/*chip=*/0);
```
krystian-hebel, line 37:
```suggestion
	const struct powerbus_cfg *pb_cfg = powerbus_cfg(chip);
```
Unless there is a reason to hardcode it?

krystian-hebel (approved):

PR #94, opened 2022-02-13
Start OCC for the second CPU

SergiiDmytruk:
Apart from CPU-specific SCOM accesses, this re-organizes 21.1 to start multiple OCCs.

### DIFF
--- a/src/include/cpu/power/occ.h
+++ b/src/include/cpu/power/occ.h
@@ -5,14 +5,14 @@
 
 #include <stdint.h>
 
-void writeOCCSRAM(uint32_t address, uint64_t *buffer, size_t data_length);
-void readOCCSRAM(uint32_t address, uint64_t *buffer, size_t data_length);
-void write_occ_command(uint64_t write_data);
-void clear_occ_special_wakeups(uint64_t cores);
-void special_occ_wakeup_disable(uint64_t cores);
-void occ_start_from_mem(void);
+void writeOCCSRAM(uint8_t chip, uint32_t address, uint64_t *buffer, size_t data_length);
+void readOCCSRAM(uint8_t chip, uint32_t address, uint64_t *buffer, size_t data_length);
+void write_occ_command(uint8_t chip, uint64_t write_data);
+void clear_occ_special_wakeups(uint8_t chip, uint64_t cores);
+void special_occ_wakeup_disable(uint8_t chip, uint64_t cores);
+void occ_start_from_mem(uint8_t chip);
 
-void pm_occ_fir_init(void);
-void pm_pba_fir_init(void);
+void pm_occ_fir_init(uint8_t chip);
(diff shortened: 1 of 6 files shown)


krystian-hebel:
@SergiiDmytruk please rebase it, for some reason GitHub shows lines that are already changed in `homer_2nd_cpu`

SergiiDmytruk:
By the way, "OCC" partition is still read twice (costing us an extra 400ms) only because I'm not sure what would be a good new place for "HCODE". Looks like can use common OCC area, any better place for it (maybe just use `malloc()`)?

krystian-hebel:
`malloc()` would be good, but it may require changing linker options, it only has 1MB of heap by default. 400ms may be worth it.

krystian-hebel:
What about WOF, is it also loaded twice?

SergiiDmytruk:
> What about WOF, is it also loaded twice?

We don't read it all, only readat header, mmap list of entries and then mmap entry that we need.

SergiiDmytruk:
Tried it in d653b558e5026ff9dda61a7647177538deae2c74, but something is wrong with FIT:
```
CBFS: Found 'fallback/payload' @0x1f8c0 size 0x17575d in mcache @0xffefeb7c
FIT: Examine payload fallback/payload
FIT: Loading FIT from 0xf8380000
FIT: Image fdt-1 has 9351 bytes.
FIT: Image kernel has 1518576 bytes.
FIT: Compat preference (lowest to highest priority) : raptor-cs,talos-ii
FIT: config conf-1 (default), kernel kernel, fdt fdt-1, compat ibm,powernv ibm,p9-openbmc rcs,talos
FIT: No match, choosing default
```

krystian-hebel (changes requested):

SergiiDmytruk:
If I drop `free()` in 45bb540bc739fea5394af54f4729471ace450816, it works. If I keep `free()`, but `memset` zeros, it works. If I keep `free()` and `memset` non-zeros, it doesn't. Some FIT/DT code uses uninitialized memory. @krystian-hebel, worth tracking down this memory issue? Maybe can find a commit that fixes it upstream, otherwise the bug is still there.

SergiiDmytruk:
`dt_find_node()` didn't zero-initialize nodes it created, the bug is present upstream as well. It's just hard to notice when there are pretty much no `free()` calls, all memory is zeroed by default and special conditions must be met for a block to be reused.

After that fix reading OCC once works as expected.

SergiiDmytruk:
Measured time it takes `wof_init()` to execute, just in case it's a lot: ~53ms for each chip. The delay comes from `rdev_mmap()` in `wof_extract()`, WOF entries might match for us, but not in general, so caching is probably unjustified (especially when compared to PNOR caching of MVPDs, which should save many seconds).

PR #96, opened 2022-02-15
ec/clevo/it5570/acpi/ucsi.asl: Disable the UCSI device

mkopec:
The UCSI device is not functional at the moment, possibly because of bugs in EC firmware. Disable it for now until we find a solution, so that no errors are displayed in the OS.


mkopec:
This breaks USB3 functionality on the USB4 port. Video and power still works, but no data. Closing this.

PR #97, opened 2022-02-15
Configure TOD for up to two CPUs

SergiiDmytruk:
Update isteps 18.11 and 18.12 to also work with 2 CPUs.


krystian-hebel (changes requested):
Few changes + rebase

krystian-hebel (approved):

PR #106, opened 2022-02-23
Switch away from FIT payload

SergiiDmytruk:
This makes skiboot an ELF payload.

Two DTB files are prepared for two configurations and included into CBFS during build. In sources 2-CPU version incudes 1-CPU file, so it should be easy to maintain. Can switch to DT overlays for greater flexibility and saving some space, code supports them and `dtc` probably does too.


SergiiDmytruk:
@krystian-hebel Copied two functions from `src/lib/fit.c`. Is it worth moving them to a separate conditionally-compiled source file and enable it for `SUPPORT_FIT_PAYLOAD` and `POWER9`? Or maybe there is another way of doing this.

src/soc/ibm/power9/chip.c
```diff
@@ -254,105 +280,88 @@ static void fill_cpu_node(struct device_tree *tree,
...
+	struct mem_map *map = arg;
 
-	if (size_cells != 2 && size_cells != 1)
-		die("'#size_cells' property in 'memory' has invalid value (%d)\n",
-		    size_cells);
+	/*
+	 * Kernel likes its available memory areas at least 1MB
+	 * aligned, let's trim the regions such that unaligned padding
+	 * is added to reserved memory.
+	 */
```
krystian-hebel, line 312: This probably applies to some embedded ARM platform, skiboot doesn't mind such regions. This whole function can be simplified to "if RAM then add to mem, else to reserved", without splicing ranges.

krystian-hebel: Actually, this may work different than Hostboot, where "memory" nodes reported all RAM, reserved or not. Reading `MCS_MCFGP*` may be the only correct option. `bootmem_walk_os_mem` would be used just for reserved regions.

src/soc/ibm/power9/chip.c
```diff
@@ -454,22 +463,80 @@ static void add_dimm_sensor_nodes(struct device_tree *tree, uint8_t chips)
...
+
+	/* Expose board ID, SKU ID, and RAM code to payload.*/
+	if (board_id() != UNDEFINED_STRAPPING_ID)
+		dt_add_u32_prop(coreboot_node, "board-id", board_id());
+
+	if (sku_id() != UNDEFINED_STRAPPING_ID)
+		dt_add_u32_prop(coreboot_node, "sku-id", sku_id());
+
+	if (ram_code() != UNDEFINED_STRAPPING_ID)
+		dt_add_u32_prop(coreboot_node, "ram-code", ram_code());
```
krystian-hebel, line 523: These are not used by our code and may be removed.

(There are some other platform identification fields in DT that will have to be added eventually, but not now)

krystian-hebel (changes requested):
This confirms what [I wrote on the mailing list](https://mail.coreboot.org/hyperkitty/list/coreboot@coreboot.org/message/42ZZNOZKFTRV6HVSDZP4EFJM5TCMWD4D/) about choosing FIT :slightly_smiling_face: 

I added some comments about stripping unnecessary parts from that code. Memory ranges walk will have to be significantly changed, but "coreboot" node is mostly the same as for FIT. Doesn't make sense to create a separate file just for that, though. Moving it to `device_tree.c` could help, but I don't know if upstream would be OK with it.

src/soc/ibm/power9/chip.c
```diff
@@ -454,22 +442,69 @@ static void add_dimm_sensor_nodes(struct device_tree *tree, uint8_t chips)
 	}
 }
 
+/**
+ * Add coreboot tables, CBMEM information and optional board specific strapping
+ * IDs to the device tree.
```
krystian-hebel, line 447:
```suggestion
 * Add coreboot tables, CBMEM information to the device tree.
```

src/mainboard/raptor-cs/talos-2/Makefile.inc
```diff
@@ -2,3 +2,20 @@
...
+
+$(obj)/$(TWOCPU_DTB): $(obj)/$(ONECPU_DTB)
+
+cbfs-files-y += $(ONECPU_DTB)
+$(ONECPU_DTB)-file := $(obj)/$(ONECPU_DTB)
+$(ONECPU_DTB)-type := raw
+
+cbfs-files-y += $(TWOCPU_DTB)
+$(TWOCPU_DTB)-file := $(obj)/$(TWOCPU_DTB)
+$(TWOCPU_DTB)-type := raw
```
krystian-hebel, line 21: Since DTB files are required by code in `soc/ibm/power9`, these lines should be added to Makefile in that directory instead. Files themself can (and should) be left in mainboard directory.

krystian-hebel (changes requested):
Additionally, `ARCH_PPC64` can be now removed from `config FIT_PAYLOAD` in `payloads/Kconfig`

EDIT: and `PAYLOAD_FIT_SUPPORT`, and probably `PAYLOAD_SKIBOOT` in that file can be removed from list of non-compressable payloads.

SergiiDmytruk:
Also removed 6 sensors from `2-cpus.dts` which were previously removed in 487d707, forgot to return to this after a rebase.

src/soc/ibm/power9/chip.c
```diff
@@ -745,10 +749,43 @@ static void activate_slave_cores(uint8_t chip)
...
+	void *fdt;
+	const char *dtb_file;
+
+	assert(chips == 0x01 || chips == 0x03);
+
+	dtb_file = (chips == 0x01 ? "1-cpu.dtb" : "2-cpus.dtb");
+	fdt = load_fdt(dtb_file, chips);
+
+	/* See asm/head.S in skiboot where fdt_entry starts at offset 0x10 */
+	prog_set_entry(prog, (void *)0x10, fdt);
```
krystian-hebel, line 787: In case payload is loaded to a different address:
```suggestion
	prog_set_entry(prog, prog_start(prog) + 0x10, fdt);
```

krystian-hebel (approved):

PR #112, opened 2022-03-07
.github/workflows/build.yml: run workflow on tag push

mkopec:


macpijan:
GitHub Actions with self-hsoted runners does not work too reliable for us. We need to come up with another approach for running CIs here.

PR #116, opened 2022-03-13
ASUS KGPE-D16: Fixed bugs in DQS Timing

mrothfuss:
Fixed logic for detecting passing read and write DQS delays
Force discard of negative DQS recovery delays


krystian-hebel:
@mrothfuss I've checked this PR against BKDG and it definitely is an improvement over old code. We haven't been able to notice any changes because our DIMMs didn't fall into path that is changed here, but based on your testing I'm willing to approve this PR.

However, I'm going to ask you to sign the commit (both Signed-of-by and digital signature) since these are our rules, even though those aren't written in publicly available documentation yet (@pietrushnic @macpijan CC).

macpijan:
We can do that as well. It is a very common model in accepting community contributions, that a maintainer cherry-picks patches and applies to a branch. You can do that as well, and as a result, it should be signed with GPG.

mrothfuss:
I retroactively gpg signed the commit (d5ddff280e014d64d3850bc8d5a6d1fd592c94ec), not sure about the signed-off-by part though.

PR #123, opened 2022-03-16
southbridge/amd/sb700: Add missing PCI bridge initialization and fix SB MMIO init

miczyg1:

### DIFF
--- a/src/southbridge/amd/sb700/pci.c
+++ b/src/southbridge/amd/sb700/pci.c
@@ -1,0 +1,125 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+
+#include <device/device.h>
+#include <device/pci.h>
+#include <device/pci_ids.h>
+#include <device/pci_ops.h>
+#include "sb700.h"
+
+static void pci_init(struct device *dev)
+{
+	u32 dword;
+	u16 word;
+	u8 byte;
+
+	/*
+	 * RPR 5.1 Enables the PCI-bridge subtractive decode
+	 * This setting is strongly recommended since it supports some legacy
+	 * PCI add-on cards,such as BIOS debug cards
+	 */
+	byte = pci_read_config8(dev, 0x4B);
+	byte |= 1 << 7;
+	pci_write_config8(dev, 0x4B, byte);
+	byte = pci_read_config8(dev, 0x40);
+	byte |= 1 << 5;
+	pci_write_config8(dev, 0x40, byte);
+
+	/*
+	 * RPR5.2 PCI-bridge upstream dual address window
+	 * This setting is applicable if the system memory is more than 4GB,
+	 * and the PCI device can support dual address access
+	 */
+	byte = pci_read_config8(dev, 0x50);
+	byte |= 1 << 0;
+	pci_write_config8(dev, 0x50, byte);
+
+	/* RPR 5.3 PCI bus 64-byte DMA read access */
+	/* Enhance the PCI bus DMA performance */
+	byte = pci_read_config8(dev, 0x4B);
+	byte |= 1 << 4;
+	pci_write_config8(dev, 0x4B, byte);
+
+	/* RPR 5.4 Enables the PCIB writes to be cacheline aligned. */
+	/* The size of the writes will be set in the Cacheline Register */
+	byte = pci_read_config8(dev, 0x40);
+	byte |= 1 << 1;
+	pci_write_config8(dev, 0x40, byte);
+
+	/*
+	 * RPR 5.5 Enables the PCIB to retain ownership of the bus on the
+	 * Primary side and on the Secondary side when GNT# is deasserted
+	 */
+	pci_write_config8(dev, 0x0D, 0x40);
+	pci_write_config8(dev, 0x1B, 0x40);
+
+	/*
+	 * RPR 5.6 Enable the command matching checking function on
+	 * "Memory Read" & "Memory Read Line" commands
+	 */
+	byte = pci_read_config8(dev, 0x4B);
+	byte |= 1 << 6;
+	pci_write_config8(dev, 0x4B, byte);
+
+	/*
+	 * RPR 5.7 When enabled, the PCI arbiter checks for the Bus Idle
+	 * before asserting GNT#
+	 */
+	byte = pci_read_config8(dev, 0x4B);
+	byte |= 1 << 0;
+	pci_write_config8(dev, 0x4B, byte);
+
+	/* RPR 5.8 Adjusts the GNT# de-assertion time */
+	word = pci_read_config16(dev, 0x64);
+	word |= 1 << 12;
+	pci_write_config16(dev, 0x64, word);
+
+	/* RPR 5.9 Fast Back to Back transactions support */
+	byte = pci_read_config8(dev, 0x48);
+	byte |= 1 << 2;
+	/* pci_write_config8(dev, 0x48, byte); */
+
+	/* RPR 5.10 Enable Lock Operation */
+	/* byte = pci_read_config8(dev, 0x48); */
+	byte |= 1 << 3;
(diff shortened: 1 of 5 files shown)


src/northbridge/amd/amdfam10/northbridge.c
```diff
@@ -536,6 +537,16 @@ static void amdfam10_create_vga_resource(struct device *dev, unsigned int nodeid
...
 	printk(BIOS_DEBUG, "VGA: %s (aka node %d) link %d has VGA device\n", dev_path(dev), nodeid, link->link_num);
 	set_vga_enable_reg(nodeid, link->link_num);
+
+
+	/* Redirect VGA memory access to MMIO This signals the Family 10h
+	 * resource parser to add a new MMIO mapping to the Range 11 MMIO
+	 * control registers (starting at F1x1B8), and also reserves the
+	 * resource in the E820 map.
+	 */
+	res = amdfam10_assign_new_mmio_res(0xa0000, 0x20000);
```
krystian-hebel, line 547: `mmio_resource()` which is already present in `amdfam10_domain_read_resources()` should do the same for E820 map (but this would be better place for that call), and adding it to MMIO control registers shouldn't be necessary because `set_vga_enable_reg()` enables this routing in different register. Also, the comment above does not match the code, it doesn't use `F1x1B8`.

src/southbridge/amd/sb700/sm.c
```diff
@@ -310,6 +310,13 @@ static void sm_init(struct device *dev)
 
 		byte |= 1 << 3;
 		pci_write_config8(dev, 0x43, byte);
+
+		/* Enable southbridge MMIO decode */
+		dword = pci_read_config32(dev, SB_MMIO_CFG_REG);
+		dword &= ~(0xffffff << 8);
+		dword |= SB_MMIO_BASE_ADDRESS;
+		dword |= 0x1;
+		pci_write_config32(dev, SB_MMIO_CFG_REG, dword);
```
krystian-hebel, line 319: Does it have to be hardcoded? Part of resource allocator v3 abuse was that there was too many hardcoded resources that were in fact configurable, like this one. Were devices under southbridge accessible during `read_resource` phase before this change?

krystian-hebel: Actually, what is the use of setting BAR if it isn't enabled? Only IO is, MMIO enable is bit 1.

krystian-hebel (changes requested):

src/southbridge/amd/sb700/sm.c
miczyg1: I have changed the address to 0xfed80000 like other AMD SBs do. Also these enable bits does not seem to work as we think. At least this is what I noticed. Mayeb it needs further checks.

miczyg1:
@krystian-hebel what are we doing with this PR?

src/southbridge/amd/sb700/sm.c
```diff
@@ -310,6 +310,13 @@ static void sm_init(struct device *dev)
 
 		byte |= 1 << 3;
 		pci_write_config8(dev, 0x43, byte);
+
+		/* Enable southbridge MMIO decode */
+		dword = pci_read_config32(dev, SB_MMIO_CFG_REG);
+		dword &= ~(0xffffff << 8);
+		dword |= SB_MMIO_BASE_ADDRESS;
+		dword |= 0x1;
```
miczyg1, line 318:
```suggestion
		dword |= 0x3;
```

src/southbridge/amd/sb700/sm.c
miczyg1: Should it be ORed with 0x2 instead?

miczyg1: Not sure why, this was in the original code from 4.11. I may clean it up.

miczyg1: @krystian-hebel I hardcoded it to 0xfed80000 because of the and commonblock for this mmio, which can be potentially used.

src/northbridge/amd/amdfam10/northbridge.c
miczyg1: Removed.

src/southbridge/amd/sb700/sm.c
miczyg1: I have moved the SB MMIO init to bootblock. Later we could select the SOC_AMD_COMMON_BLOCK_ACPIMMIO and replace all the occurrences of reimplemented SB MMIO access through I/O.

krystian-hebel (changes requested):
@miczyg1 please rebase to current develop and remove merge conflict markers added by "configs/asus_kgpe-d16: Use linear framebuffer " (probably drop whole commit?)

src/southbridge/amd/sb700/pci.c
```diff
@@ -0,0 +1,110 @@
...
+#include "sb700.h"
+
+static void pci_init(struct device *dev)
+{
+	u32 dword;
+	u16 word;
+	u8 byte;
+
+	/* RPR 5.1 Enables the PCI-bridge subtractive decode */
+	/* This setting is strongly recommended since it supports some legacy PCI add-on cards,such as BIOS debug cards */
```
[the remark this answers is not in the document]
miczyg1: Done for all long comment in the file

miczyg1:
> @miczyg1 please rebase to current develop and remove merge conflict markers added by "configs/asus_kgpe-d16: Use linear framebuffer " (probably drop whole commit?)

It is rebased on the recent develop branch. Dropped a few more irrelevant commits.

src/southbridge/amd/sb700/sm.c
```diff
@@ -475,11 +482,13 @@ static void sb700_sm_read_resources(struct device *dev)
 	 * write address in set_resources() below.
 	 */
 	res = new_resource(dev, SB_MMIO_CFG_REG);
+	res->base  = SB_MMIO_BASE_ADDRESS;
 	res->size = 0x1000;
 	res->limit = 0xffffffffUL;
 	res->align = 12;
 	res->gran = 12;
-	res->flags = IORESOURCE_MEM | IORESOURCE_ASSIGNED;
+	res->flags = IORESOURCE_MEM | IORESOURCE_FIXED | IORESOURCE_ASSIGNED |
```
krystian-hebel, line 483: Since this is fixed and stored, "Enable SB MMIO" in function below (`sb700_sm_set_resources`) is no longer necessary, please remove it.

I'm still not convinced on hardcoding the address, but if this is the case across all AMD boards, it is better to not make an exception here.

miczyg1: Isn't it necessary to just inform the resource allocator about the resource presence anyways? So that it will not allocate any other resource here,

miczyg1: Enabling SB MMIO removed from sb700_sm_set_resources

miczyg1: @krystian-hebel can we resolve this conversation and close the PR?

src/southbridge/amd/sb700/pci.c
```diff
@@ -0,0 +1,125 @@
...
+static void pci_init(struct device *dev)
+{
+	u32 dword;
+	u16 word;
+	u8 byte;
+
+	/*
+	 * RPR 5.1 Enables the PCI-bridge subtractive decode
+	 * This setting is strongly recommended since it supports some legacy
+	 * PCI add-on cards,such as BIOS debug cards
```
krystian-hebel, line 18:
```suggestion
	 * PCI add-on cards, such as BIOS debug cards
```

miczyg1:
@krystian-hebel we need an approve as well

krystian-hebel (approved):

PR #124, opened 2022-03-16
tuxedo_ibs15: build with BUILD_TIMELESS

mkopec:


macpijan:
BILD_TIMELESS is not something we want to use, as it just removes the git commit information from the build (we want to preserve it).

PR #126, opened 2022-03-17
build.sh: install vboot utils locally

mkopec:
Fixes building vboot testing binaries in CI.


mkopec:
@macpijan this time CI passes and vboot binaries get built correctly

PR #127, opened 2022-03-17
build vboot test images in CI

mkopec:


macpijan:
GitHub Actions with self-hsoted runners does not work too reliable for us. We need to come up with another approach for running CIs here.

macpijan:
GitHub Actions with self-hsoted runners does not work too reliable for us. We need to come up with another approach for running CIs here.

PR #136, opened 2022-03-25
Novacustom ns5x/rgb keyboard

mkopec:
TODO: Call `KBUP ()` in a more appropriate place


src/ec/clevo/it5570/acpi/rgb_kb.asl
```diff
@@ -0,0 +1,105 @@
+Name (KBEN, 0x01) // Keyboard Backlight Enable
```
miczyg1, line 1: license header missing. It would be good to issue `make test-lint` when issuing a PR

miczyg1 (changes requested):
Looks, good, just need the license header in the added file

miczyg1:
@mkopec one concern: doesn't it conflict with the single color backlight on NV41? We should think about two platforms when developing anything in the ec directory.

mkopec:
@miczyg1 I tested this on an nv41, everything regarding white keyboard backlight still works as it should. No side effects after closing the lid or sleeping.

miczyg1 (approved):

PR #140, opened 2022-04-04
NOT FOR MERGE: dasharo/common base

pietrushnic:
This branch should stay as dasharo/common-base to provide common improvements for all Dasharo open-source firmware suppported hardware. We should consider versioning and update process for Dasharo Common Base as it is foundation for all future hardware support.


payloads/external/tianocore/Kconfig
```diff
@@ -186,4 +186,28 @@ config TIANOCORE_USE_8254_TIMER
 
 endif
 
+config CONFIG_TIANOCORE_ENABLE_IPXE
```
miczyg1, line 189: We laso need TIANOCORE_BOOTSPLASH_FILE and TIANOCORE_HAVE_EFI_SHELL to be outside of the `if TIANOCORE_UEFIPAYLOAD` above

pietrushnic: @miczyg1 I did that fix, but I have no idea what is the logic behind that. Even if these options are payload independent, then those should be under `UEFIPAYLOAD || COREBOOTPAYLOD` since enabling those without any payload doesn't make sense.

pietrushnic:
We continue in #194

PR #145, opened 2022-04-06
mainboard/msi/ms7d25: Add buildable mainboard tree

miczyg1:


pietrushnic:
@miczyg1 why we not use upstream first? Or maybe we use, but I don't see the refernce?

miczyg1:
We can start with upstreaming if something at least works. We haven't yet tested any coreboot PR on hardware. Marking patches as WIP also won't help in getting more reviews right now.

PR #147, opened 2022-04-11
Raptor cs talos 2/rel v0.5.0

krystian-hebel:


dasharo/raptor-cs_talos-2_v0.5.0.yaml
```diff
@@ -0,0 +1,96 @@
...
+    changed:
+    - TBD
+    hardware_matrix:
+      url: https://docs.dasharo.com/variants/talos_2/hardware-matrix/
+    known_issues:
+    - TBD
+    removed:
+    - TBD
+    results:
+      failed: TBD
```
macpijan, line 89: @krystian-hebel We can adjust test results and generate final HTml file

PR #149, opened 2022-04-12
Msi ms7d25/rel v0.1.0

miczyg1:


dasharo/project_config.yaml
```diff
@@ -0,0 +1,36 @@
+# Dasharo project configuration data
+
+dasharo:
+  platform: msi_ms7d25
```
macpijan, line 4: @miczyg1 please use the latest file format (e.g. the newsletter url link)

miczyg1: Yes, I haven't filled all the necessary information back then. Now it is fixed. asciinema still missing.

PR #162, opened 2022-04-27
Fixes for single CPU setup

SergiiDmytruk:
Fixes some issues in Dasharo/dasharo-issues#80 (GitHub won't close the issue on merging this PR).

Tested by making `fsi_get_present_chips()` return `0x01`. Potentially, there can be other issues, but Linux boots fine with hard-coded mask after these fixes.

### DIFF
--- a/src/soc/ibm/power9/homer.c
+++ b/src/soc/ibm/power9/homer.c
@@ -1659,7 +1659,19 @@ static void wait_for_occ_response(struct homer_st *homer, uint32_t timeout_sec,
 				  uint8_t seq_num)
 {
 	enum {
-		OCC_RSP_SAMPLE_TIME_US = 10,
+		/*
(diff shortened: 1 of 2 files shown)


krystian-hebel:
@SergiiDmytruk please rebase, otherwise LGTM. I thought we have it merged already.

SergiiDmytruk:
Rebased. By the way, I later saw that we already have an option for CPU count in `src/mainboard/raptor-cs/talos-2/Kconfig`:
```kconfig
config MAX_CPUS
	int
	default 1
```
But it's not used in the code.

krystian-hebel:
Nice catch. I guess `MAX_CPUS` could be used instead of https://github.com/Dasharo/coreboot/blob/raptor-cs_talos-2/develop/src/include/cpu/power/proc.h#L8

PR #151, opened 2022-04-12
Add options to enable vboot

IgorBagnucki:
Signed-off-by: Igor Bagnucki <igor.bagnucki@3mdeb.com>

PR #166, opened 2022-05-04
Use SPI flash lock for vboot

SergiiDmytruk:
Will lock `WP_RO` region of the flash on startup using SPI controller.


SergiiDmytruk:
It was enabled in #151, this just enables flash lockdown. I was actually thinking of adding this to `src/mainboard/protectli/vault_cml/Kconfig` under `config VBOOT`, but didn't. That's the place for it, right?

miczyg1:
NVM, the CONFIG_VBOOT is in the minimal config file (somehow I missed it)

miczyg1 (approved):

PR #175, opened 2022-05-18
ec/clevo/it5570: store keyboard backlight in CMOS

mkopec:
Add persistent keyboard implementation for laptops with RGB keyboards like NS5x.

Store the keyboard settings in CMOS. Add defaults: Enabled, White color, 20% brightness.

### DIFF
--- a/src/ec/clevo/it5570/acpi/cmos.asl
+++ b/src/ec/clevo/it5570/acpi/cmos.asl
@@ -1,0 +1,18 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+
+OperationRegion (CMS1, SystemIO, 0x72, 0x2)
+Field (CMS1, ByteAcc, NoLock, Preserve)
+{
+	IND1, 8,
+	DAT1, 8,
+}
+
+IndexField (IND1, DAT1, ByteAcc, NoLock, Preserve)
+{
+	KBBR, 8,		// Keyboard Backlight Brightness
+	KBEN, 8,		// Keyboard Backlight State
+	KBCR, 8,		// Keyboard Backlight Color Red
+	KBCG, 8,		// Keyboard Backlight Color Green
+	KBCB, 8,		// Keyboard Backlight Color Blue
+	KBPS, 8,		// Keyboard Backlight Preset
+}
--- a/src/ec/clevo/it5570/acpi/ec.asl
+++ b/src/ec/clevo/it5570/acpi/ec.asl
@@ -30,6 +30,7 @@ Device (EC0)
 	})
 
 	#include "ec_ram.asl"
+	#include "cmos.asl"
 
 	Name (ECOK, Zero)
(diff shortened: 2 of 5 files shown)


miczyg1:
Question: should coreboot restore these settings from CMOS to EC on each boot? I recall customers complaining on LED settings were lost after reset/power cycle

mkopec:
@miczyg1 Yes, and with this PR the settings are restored during `_REG`

src/ec/clevo/it5570/acpi/cmos.asl
```diff
@@ -0,0 +1,19 @@
...
+OperationRegion (CMS1, SystemIO, 0x72, 0x2)
+Field (CMS1, ByteAcc, NoLock, Preserve)
+{
+	IND1, 8,
+	DAT1, 8,
+}
+
+IndexField (IND1, DAT1, ByteAcc, NoLock, Preserve)
+{
+	Offset (0x80),
```
miczyg1, line 12: This is baaad. The first CMOS bank 0x70/0x71 has 128 bytes (0x00-0x7f). Any CMOS access to index above 128 is actually subtracted by 128 and CMOS I/O port pair used should be 0x72/0x73. Soooo... If these settings are starting at bit 1024 (the 128th byte) that means this is offset 0 of the 2nd CMOS bank (0x72/0x73).

miczyg1 (changes requested):

miczyg1:
> @miczyg1 Yes, and with this PR the settings are restored during `_REG`

I meant earlier, before ACPI interpreters kick in. E.g. one enters the Setup menu, the keyboard backlight will not be restored at this point.

mkopec:
@miczyg1 Right, we can restore the settings in coreboot instead, that does make sense

miczyg1:
> @miczyg1 Right, we can restore the settings in coreboot instead, that does make sense

If _REG does it, that's also fine. There should be no harm in repeating the operation (I hope so).

src/ec/clevo/it5570/acpi/cmos.asl
mkopec: Ack, copy-paste error from https://github.com/Dasharo/coreboot/blob/95af254e116802490481c6293ae752f5b61e561b/src/ec/starlabs/merlin/acpi/cmos.asl#L19-L33

mkopec:
Sending the FCMDs for keyboard from coreboot doesn't work. Maybe the EC needs to be in ACPI mode first?

miczyg1:
> Sending the FCMDs for keyboard from coreboot doesn't work. Maybe the EC needs to be in ACPI mode first?

Could be. If it doesn't work, let's leave it.

src/ec/clevo/it5570/acpi/cmos.asl
mkopec: Fixed

miczyg1 (approved):

PR #178, opened 2022-05-23
Novacustom ns5x/rel 1.2.0

mkopec:
- cherrypick SMM loader refactor
- persistent keyboard backlight
- remove power limits (to use SKU defaults)
- tweak fan curve to keep the fan spinning up to 100 degrees C and avoid spinup noise
- Change BIOS vendor name to 3mdeb


mkopec:
@miczyg1 I think this is ready for review now

miczyg1 (approved):

PR #185, opened 2022-06-05
I2C TPM support

SergiiDmytruk:


src/soc/ibm/power9/i2c.c
```diff
@@ -176,6 +175,37 @@ int platform_i2c_transfer(unsigned int bus, struct i2c_msg *segment,
...
+	 * -  400 kHz - 0x0177
+	 *
+	 * Use value for 400 kHz as it is the one used by Hostboot.
+	 */
+	uint16_t bit_rate_div = 0x0177;	// 400kHz by default
+	if (bus == 2) {
+		/*
+		 * Skiboot computes the value as:
+		 *
+		 *   (((clock-frequency / bus-frequency) - 1) / 4)
```
krystian-hebel, line 199: `clock-frequency` is frequency of core? If that is the case, it may cause problems for other buses, too. This also may mean that values change after OCC is started (boot frequency from scratch registers vs. target frequency from MVPD).

0x0177 was read on our CPU, but if it depends on CPU frequency, it may even be the cause of https://github.com/Dasharo/dasharo-issues/issues/80 if RCD is programmed incorrectly.

src/soc/ibm/power9/Kconfig
```diff
@@ -22,4 +22,22 @@ config SKIP_INITIAL_ECC_SCRUB
 	  Selecting this option almost halves the time needed for memory
 	  initialization (istep 14.1).
 
+if BOARD_RAPTOR_CS_TALOS_2
+
+config TALOS_2_INFINEON_TPM_1
```
krystian-hebel, line 27: SoC directory should be clear of mainboard names, if possible. I assume this is done to hide TPM option for QEMU?

src/mainboard/raptor-cs/talos-2/Kconfig
```diff
@@ -27,6 +27,23 @@ config BOARD_SPECIFIC_OPTIONS
 	select IPMI_BT_ROMSTAGE
 	select FLATTENED_DEVICE_TREE
 
+config TALOS_2_INFINEON_TPM_1
```
miczyg1, line 30: This can't be define like that. MAINBOARD_HAS_TPM1 should be selected by BOARD_SPECIFIC_OPTIONS and then TPM_MEASURED_BOOT and MAINBOARD_HAS_TPM1 should be selectable in the menuconfig or saved in miniconfig file.

src/mainboard/raptor-cs/talos-2/Kconfig
```diff
@@ -27,6 +27,23 @@ config BOARD_SPECIFIC_OPTIONS
...
+	bool "I2C SLB9635TT or SLB9645TT or another TPM1 chip compatible with SLB9635TT"
+	default n
+	select TPM_MEASURED_BOOT # needed for TCPA log
+	select MAINBOARD_HAS_I2C_TPM_GENERIC
+	select MAINBOARD_HAS_TPM1
+
+config DRIVER_TPM_I2C_BUS
+	hex "I2C TPM chip bus"
+	default 0x02
+	depends on TALOS_2_INFINEON_TPM_1
```
miczyg1, line 40: No need to depend on that I guess

src/mainboard/raptor-cs/talos-2/Kconfig
```diff
@@ -27,6 +27,23 @@ config BOARD_SPECIFIC_OPTIONS
...
+
+config DRIVER_TPM_I2C_ADDR
+	hex "I2C TPM chip address"
+	default 0x20
+	depends on TALOS_2_INFINEON_TPM_1
```
miczyg1, line 45: a/a

src/soc/ibm/power9/chip.c
```diff
@@ -488,6 +490,42 @@ static void add_cb_fdt_data(struct device_tree *tree)
 	dt_add_reg_prop(coreboot_node, reg_addrs, reg_sizes, 2, addr_cells, size_cells);
 }
 
+static void add_tpm_node(struct device_tree *tree)
+{
+#if CONFIG(TALOS_2_INFINEON_TPM_1)
```
miczyg1, line 495: This should depend on whether TPM is really selected in the menu or TPM is detected.
Furthermore I would do:

```
if (!CONFIG(SOEMTHING))
    return;
```

src/mainboard/raptor-cs/talos-2/Kconfig
SergiiDmytruk: This is to not show these fields unless TPM support is enabled.

src/soc/ibm/power9/chip.c
SergiiDmytruk: > This should depend on whether TPM is really selected in the menu or TPM is detected.

If it wasn't selected, the code for detecting TPM wasn't compiled.

> Furthermore I would do:

Then it won't link when `TALOS_2_INFINEON_TPM_1` is off.

src/mainboard/raptor-cs/talos-2/Kconfig
SergiiDmytruk: > MAINBOARD_HAS_TPM1 should be selected by BOARD_SPECIFIC_OPTIONS and then TPM_MEASURED_BOOT and MAINBOARD_HAS_TPM1 should be selectable in the menuconfig or saved in miniconfig file.

I thought that including it in `BOARD_SPECIFIC_OPTIONS` will enable it unconditionally, no? Or do you have a typo somewhere? I didn't include `MAINBOARD_HAS_TPM1` in `BOARD_SPECIFIC_OPTIONS` because mainboard can work with TPM2 as well, or should it be selected conditionally if `MAINBOARD_HAS_TPM1` is on?

src/mainboard/raptor-cs/talos-2/Kconfig
miczyg1: Ahh right. But then is it possible to have TPM on different I2C buses?

src/mainboard/raptor-cs/talos-2/Kconfig
```diff
@@ -27,6 +27,23 @@ config BOARD_SPECIFIC_OPTIONS
 	select IPMI_BT_ROMSTAGE
 	select FLATTENED_DEVICE_TREE
 
+config TALOS_2_INFINEON_TPM_1
+	bool "I2C SLB9635TT or SLB9645TT or another TPM1 chip compatible with SLB9635TT"
```
krystian-hebel, line 31: This doesn't fit in menuconfig launched in standard 80x25 shell window
```suggestion
	bool "I2C TPM1 chip compatible with SLB9635TT"
```

src/mainboard/raptor-cs/talos-2/Kconfig
krystian-hebel: This may be undiscovered territory, most (all?) of I2C TPMs used by coreboot are probably soldered. AFAICT only `mainboard/google/gru` has support for different TPM families. I suggest something similar to that.

src/soc/ibm/power9/chip.c
krystian-hebel: It would link if it were inside this function. I think what miczyg said was to use normal C `if` instead of precompiler's `#if`, and reverse the condition checked.

src/soc/ibm/power9/i2c.c
```diff
@@ -260,7 +273,7 @@ int platform_i2c_transfer(unsigned int bus, struct i2c_msg *segment,
 
 	}
 
-	return bytes_transfered;
+	return 0;
```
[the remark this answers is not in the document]
krystian-hebel: I'm marking it as resolved. It should be consistent and documented (at least by a comment in header file), but this is outside of scope of this task, and we can't really change it without having to change other platform's code.

krystian-hebel (changes requested):
Code works in general. However, this is not the end of implementation, there are pieces of logic that are not measured (e.g. other PNOR partitions used), which will have to be done in following task. It also looks suspicious to me that only PCR2 and PCR10 are used, but maybe that's expected, I haven't played with TPM1.2 too much. We will also have to consider root of trust at some point.

For now there are still some small changes requested in comments and if those are resolved we can merge this PR.

src/soc/ibm/power9/i2c.c
SergiiDmytruk: It seems to come from skiboot. Clock frequency of i2cm is xscom frequency divided by 4. And for bus-frequency:
```c
		/*
		 * use the clock frequency as the bus frequency until we
		 * have actual devices on the bus. Adding a device will
		 * reduce the frequency to something that all devices
		 * can tolerate.
		 */
```
Although I don't see where it's updated.

SergiiDmytruk:
I didn't mean this to be a complete TPM support, only that this makes at least this one chip work.

src/soc/ibm/power9/chip.c
SergiiDmytruk: It can't link as is, because `tis_name` is not defined anywhere then, but I guess TPM code can be linked and unused. Might need to add a way to check whether TPM is present, it seems to be static at the moment. Or can leave it as is because skiboot disables TPM in DT if there is an issue with it and the next PR will remove
```c
dt_add_string_prop(tpm, "status", "okay");
```

krystian-hebel: Right, I missed it.

TPM presence is static because it is often used together with measured/verified boot. Having it detectable at run time would make attacks as simple as disconnecting the TPM.

src/soc/ibm/power9/i2c.c
krystian-hebel: Ack. I expected it to come from machine XML but it doesn't seem to be the case.

PR #190, opened 2022-06-09
Asus kgpe d16/vga fix

krystian-hebel:


src/southbridge/amd/sr5650/bootblock.c
```diff
@@ -19,6 +47,7 @@ void sr5650_disable_pcie_bridge(void)
 	mask |= (1 << 20); /*GPP3b*/
 	reg = mask;
 	set_nbmisc_enable_bits(nb_dev, 0x0c, mask, reg);
+	enable_pcie_bar3();
```
miczyg1, line 50: This function is public so maybe lets' move it to bootblock_early_southbridge_init in sb700/bootblock as other sr5650 into functions are there

miczyg1 (changes requested):

src/northbridge/amd/amdfam10/northbridge.c
```diff
@@ -536,6 +536,8 @@ static void amdfam10_create_vga_resource(struct device *dev, unsigned int nodeid
 
 	printk(BIOS_DEBUG, "VGA: %s (aka node %d) link %d has VGA device\n", dev_path(dev), nodeid, link->link_num);
 	set_vga_enable_reg(nodeid, link->link_num);
+
+	reserved_ram_resource(dev, 6, 0xa0000 >> 10, 0x20000 >> 10);
```
miczyg1, line 540: I just removed it in the other PR :) So what and why exactly do we do here?

src/southbridge/amd/sb700/sm.c
```diff
@@ -498,7 +498,7 @@ static void sb700_sm_set_resources(struct device *dev)
 	 * as soon as possible.
 	 */
 	byte = pci_read_config8(dev, SB_MMIO_CFG_REG);
```
miczyg1, line 500: This has been moved to bootblock in the other PR.

src/southbridge/amd/sr5650/bootblock.c
krystian-hebel: Done

src/northbridge/amd/amdfam10/northbridge.c
krystian-hebel: This is added for case with dGPU and disabled Aspeed - in that case this region wouldn't be reserved.

src/southbridge/amd/sb700/sm.c
krystian-hebel: AFAICT there is no need to do this in bootblock.

miczyg1: If we would like to use amdblocks/acpimmio it would be needed. But ok, let's not mess with it here.

miczyg1 (approved):

PR #191, opened 2022-06-10
Protectli vault cml/prod variant meminit

mkopec:
Fixes for meminit for the production version of the VP4620.

Not yet tested on the preproduction variant - may have issues with memory init there. Need to test (in case the manufacturer switches back to the previous CML stepping)


mkopec:
Enabled the PTT fTPM and disabled dTPM support/ Pulled in some of @miczyg1's fixes for PTT. This fixes Vboot on VP4620 on boards without discrete TPMs

src/mainboard/protectli/vault_cml/devicetree.cb
```diff
@@ -53,10 +53,10 @@ chip soc/intel/cannonlake
 	register "PchPmSlpSusMinAssert" = "2"	# 500ms
 	register "PchPmSlpAMinAssert"   = "4"	# 2s
 
-	register "tcc_offset" = "5" # TCC of 95C	
-	
+	register "tcc_offset" = "5" # TCC of 95C
```
miczyg1, line 56: @macpijan  Any requests about this one? I recall Protectli wanted 75 degrees last time (but for FW6?)

src/security/vboot/Kconfig
```diff
@@ -95,6 +95,16 @@ config VBOOT_STARTS_IN_ROMSTAGE
 	  memory initialization). This implies that the vboot work buffer is
 	  in CBMEM from the start and doesn't need to be reserved in memlayout.
 
+config VBOOT_NO_TPM
+	bool
+	default y if !TPM
+	default n
+	select VBOOT_MOCK_SECDATA
```
miczyg1, line 102: @mkopec if HAVE_INTEL_PTT is selected, the mock is already enabled. Do we need this change too?

miczyg1 (changes requested):
I think we are also missing this: https://review.coreboot.org/c/coreboot/+/64221
if we are pulling my fixes.

src/mainboard/protectli/vault_cml/devicetree.cb
macpijan: For FW6 the expectation was to match the settings from the stock FW - so the 75 deg, Maybe as a standard we should follow the same path, by checking how is it set on the stock FW first.

src/security/vboot/Kconfig
mkopec: VBOOT_NO_TPM is also valid without HAVE_INTEL_PTT, for example when a dTPM may not always be present. We still need to mock secdata in that case

miczyg1: VBOOT_NO_TPM is not useful if HAVE_INTEL_PTT is selected. Note HAVE_INTEL_PTT does not say that you want to use CRB TPM, you still need to select MAINBOARD_HAS_CRB_TPM. You can select HAVE_INTEL_PTT, but don't have to.

miczyg1: What I am trying to say is that, HAVE_INTEL_PTT will select mocking the secdata already. Adding VBOOT_NO_TPM does not change anything now...

mkopec: It affects these lines:

- https://github.com/Dasharo/coreboot/blob/f056a9d7b857bead3ef5e4f75397208618010b72/src/drivers/tpm/Kconfig#L4
- https://github.com/Dasharo/coreboot/blob/f056a9d7b857bead3ef5e4f75397208618010b72/src/security/tpm/tspi/tspi.c#L101
- https://github.com/Dasharo/coreboot/blob/f056a9d7b857bead3ef5e4f75397208618010b72/src/security/vboot/tpm_common.h#L3

So basically it disables all TPM accesses in vboot, which isn't done when MOCK_SECDATA is selected. some tpm accesses were still being made in verstage, which will not work if:

- HAVE_LPC_TPM is selected but LPC TPM is not present
- HAVE_CRB_TPM is selected but PTT is not functional

The point of this Kconfig was to disable all TPM handling in vboot and do all TPM init in ramstage, which will be useful when we implement support for PTT / dTPM switching and PTT enable/disable via HECI which can only be done after DRAM init

miczyg1: `HAVE_CRB_TPM is selected but PTT is not functional` this will not work even in ramstage if ME is not function.
You still can have HAVE_INTEL_PTT which selects MOCK_SECDATA and choose HAVE_LPC_TPM, there is no problem with that.

> The point of this Kconfig was to disable all TPM handling in vboot and do all TPM init in ramstage, which will be useful > when we implement support for PTT / dTPM switching and PTT enable/disable via HECI which can only be done after > DRAM init

And there is no benefit from adding VBOOT_NO_TPM still, because:
1. MOCK_SECDATA will still be enabled because of HAVE_INTEL_PTT
2. MOCK_SECDATA will prevent TPM NVRAM accesses in vboot, although measured boto will still try TPM access (not sure if you have thought about it...)
3. Before you can add switching you need to unify CRB and LPC TPM drivers in coreboot. And when you already do that, tpm_init can be done whenever we want, because it will use the right driver.
4. Once you disable PTT you need to issue a global reset. After global reset, the dTPM will be functional and the unified driver from point 3 will detect which TPM it is and use proper init (either LPC or CRB depending on the state of PTT).

If something was still hanging despite HAVE_INTEL_PTT was selected, you were doing something wrong.

mkopec: > `HAVE_CRB_TPM is selected but PTT is not functional` this will not work even in ramstage if ME is not function.

It will help if it's the first time we've booted after selecting PTT in the setup menu and we haven't sent the HECI command to enable PTT yet. Without this change we will not be able to get through verstage.

> measured boot will still try TPM access (not sure if you have thought about it...)

I have, which is the point of [this change](https://github.com/Dasharo/coreboot/blob/f056a9d7b857bead3ef5e4f75397208618010b72/src/security/tpm/tspi/tspi.c#L101). With it, measurements will be stored in TCPA log until `tpm_setup` is called in ramstage.

Also note that we haven't yet decided whether PTT or dTPM will be enabled in the upcoming release. If we choose to go with dTPM only, then this change will be needed, otherwise platforms without dTPMs installed will be rendered unbootable.

src/mainboard/protectli/vault_cml/devicetree.cb
miczyg1: @mkopec would you check?

miczyg1 (approved):

src/mainboard/protectli/vault_cml/devicetree.cb
mkopec: from turbostat on vendor firmware:

```
MSR_IA32_TEMPERATURE_TARGET: 0x05640000
```

so 5 degrees is correct.

PR #192, opened 2022-06-10
Make trusted boot work in skiboot

SergiiDmytruk:
This adds an option to change TCPA log format (from coreboot-specific to TPM2 format) and updates device tree to enable trusted boot in skiboot.

Initially used log format from TPM1.2, but skiboot parsed only first entry, so changed it. Linux can handle both formats, not sure how it picks the one to use though. In case of issues with Linux, might modify skiboot to use the other format.

With this PR and corresponding changes in skiboot (https://github.com/Dasharo/skiboot/pull/1), it imports log, appends to it and updates PCRs.


src/commonlib/include/commonlib/tcpa_log_serialized.h
```diff
@@ -6,12 +6,49 @@
...
+	char name[TCPA_PCR_HASH_NAME];
+} __packed;
+
+/* TCG_PCR_EVENT2 */
+struct tcpa_entry {
+	uint32_t pcr;
+	uint32_t event_type;
+	uint32_t digest_count;
+	uint16_t digest_type;
+	uint8_t digest[TCPA_DIGEST_MAX_LENGTH];
```
krystian-hebel, line 36: This will break TCG format. `name_length` (or `EventSize`, or `eventDataSize`, depending on specification) should immediately follow last meaningful byte of digest, meaning this array will have different size based on digest algorithm. It is impossible to pack generalized form in C struct. We can either implement this with a walking pointer and append pairs of digest type and digest, or change the size of entry to SHA1 and put a warning that it is hardcoded and why. First option would be useful for TPM2.0.

src/lib/coreboot_table.c
```diff
@@ -248,7 +248,9 @@ static void add_cbmem_pointers(struct lb_header *header)
 		{CBMEM_ID_ACPI_GNVS, LB_TAG_ACPI_GNVS},
 		{CBMEM_ID_VPD, LB_TAG_VPD},
 		{CBMEM_ID_WIFI_CALIBRATION, LB_TAG_WIFI_CALIBRATION},
+#if !CONFIG(TPM_MEASURED_BOOT_SPEC_LOG)
```
krystian-hebel, line 251: Is this necessary? Code simply won't find this section if it isn't present. Talos2 doesn't have most of the rest anyway.

src/security/tpm/Makefile.inc
```diff
@@ -55,10 +55,22 @@ romstage-y += tspi/crtm.c
 ramstage-y += tspi/crtm.c
 postcar-y += tspi/crtm.c
 
+ifeq ($(TPM_MEASURED_BOOT_SPEC_LOG),y)
```
krystian-hebel, line 58: Shouldn't this be reversed?

src/soc/ibm/power9/chip.c
```diff
@@ -512,16 +513,22 @@ static void add_tpm_node(struct device_tree *tree)
...
 
-	dt_add_u64_prop(tpm, "linux,sml-base", (uintptr_t)tcpa_table + sizeof(*tcpa_table));
+	dt_add_u64_prop(tpm, "linux,sml-base", (uintptr_t)&tcpa_table->header);
 	dt_add_u32_prop(tpm, "linux,sml-size",
+			sizeof(struct tcpa_header) +
 			tcpa_table->max_entries * sizeof(struct tcpa_entry));
+
+	/* Not hard-coding into DTS-file in case will need to store key hash here */
+	sb = dt_find_node_by_path(tree, "/ibm,secureboot", NULL, NULL, 1);
+	dt_add_string_prop(sb, "compatible", "ibm,secureboot-v1-softrom");
```
krystian-hebel, line 529: Depending on securerom will make bootblock in SEEPROM difficult (HBBL has securerom appended which takes almost 9KB out of 20KB available), but we will worry about it when we get there.

src/soc/ibm/power9/chip.c
```diff
@@ -512,16 +513,22 @@ static void add_tpm_node(struct device_tree *tree)
...
+	dt_add_string_prop(sb, "hash-algo", "sha512");
```
krystian-hebel, line 530: How does Skiboot handle SHA512 on TPM1.2? Is it truncated or extended with full content of SHA512?

krystian-hebel (changes requested):

src/security/tpm/Makefile.inc
SergiiDmytruk: Of course.. it also lacks `CONFIG_` prefix which is why I didn't notice wrong condition.

src/lib/coreboot_table.c
SergiiDmytruk: It is present, but not in the format `coreboot` normally uses. I just didn't want to export table in incompatible format.

krystian-hebel: Good point. In that case maybe a new ID should be added?

src/soc/ibm/power9/chip.c
SergiiDmytruk: The `hash-algo` key was added to prevent skiboot from aborting. Without secure boot it's not used I think, but skiboot still requests the key from DT and that code fails if key is missing.

src/soc/ibm/power9/chip.c
SergiiDmytruk: The only consideration behind picking `softrom` was to involve less initialization code and get to TPM usage faster.

krystian-hebel: Makes sense, the code is compiled in anyway.

src/commonlib/include/commonlib/tcpa_log_serialized.h
SergiiDmytruk: Thanks, I missed that. I found the right specification only at the very end, because TCG has the worst reference organization I've ever seen.

src/lib/coreboot_table.c
SergiiDmytruk: Yeah, that's a better way of doing it.

krystian-hebel (approved):
Approved. I've got a feeling that upstream will want to make it useful for TPM2.0, where e.g. SHA1 is deprecated in the newest revisions of specification.

PR #197, opened 2022-06-22
Make cbmem work and print TCPA log

SergiiDmytruk:
This includes cleaned up changes from https://github.com/Dasharo/coreboot/pull/62, because they were necessary.

If https://github.com/Dasharo/skiboot/pull/2 is merged first, this PR can be changed to update skiboot payload to use that fork.

### DIFF
--- a/util/cbmem/cbmem.c
+++ b/util/cbmem/cbmem.c
@@ -20,6 +20,7 @@
 #include <commonlib/cbmem_id.h>
 #include <commonlib/timestamp_serialized.h>
 #include <commonlib/tcpa_log_serialized.h>
+#include <commonlib/tpm_log_serialized.h>
 #include <commonlib/coreboot_tables.h>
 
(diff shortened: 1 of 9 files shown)


util/cbmem/cbmem.c
```diff
@@ -384,15 +413,22 @@ static int parse_cbtable(u64 address, size_t table_size)
 			continue;
 		}
 
+		mismatched_endianness = (lbh->header_bytes >= 0x01000000);
```
krystian-hebel, line 416: This seems fragile if for any reason header grows up to >=256 bytes. Maybe compare it with `sizeof(struct lb_header)`, both normal and `bswap`-ped? This could additionally protect against different versions of this structure, although I expect the signature to change in that case.

krystian-hebel (approved):
@SergiiDmytruk approved, but please also change Skiboot revision in this PR.

PR #201, opened 2022-06-28
Common base

pietrushnic:


.gitmodules
```diff
@@ -1,63 +1,77 @@
...
 [submodule "3rdparty/intel-sec-tools"]
 	path = 3rdparty/intel-sec-tools
-	url = ../9esec-security-tooling.git
+	url = https://review.coreboot.org/9esec-security-tooling.git
 [submodule "3rdparty/stm"]
 	path = 3rdparty/stm
-	url = ../STM
+	url = https://review.coreboot.org/STM.git
 	branch = stmpe
+[submodule "util/nvidia/cbootimage"]
```
miczyg1, line 64: Duplicates

util/genbuild_h/genbuild_h.sh
```diff
@@ -66,6 +66,10 @@ printf "#define COREBOOT_VERSION %s\n" "\"$KERNELVERSION\""
 
 #See if the build is running in a git repo and the git command is available
 printf "/* timesource: $TIMESOURCE */\n"
+
+printf "#define DASHARO_VERSION \"%s\"\n" `git describe --tags --match "asus_kgpe-d16_v[0-9]*.[0-9]*.[0-9]*" | sed 's/asus\_kgpe\-d16\_//' | sed 's/\(v[0-9]\.[0-9]\.[0-9]\+\).*/\1/'`
```
miczyg1, line 70: Needs unification across other branches. OptiPlex does not use DASHARO_VERSION variables. Also these macros depend on the board name. Maybe we could get it via branch name somehow?

miczyg1 (changes requested):

util/genbuild_h/genbuild_h.sh
mkopec: This will make it more difficult to support multiple devices in a single branch which I think we want to do, eventually

.gitmodules
mkopec: Yes, not sure what happened here

pietrushnic:
@mkopec @miczyg1 I guess we rebased that multiple times and this PR is no longer valid, can we delete PR and branch?

miczyg1:
What si the point if we delete the branch? Nobody will have a chance to make use of it then. PR is not needed though

pietrushnic:
@miczyg1 shouldn't we use this https://github.com/Dasharo/coreboot/tree/common-base-rebased ? Please do cleanup of branches because its getting messy.

miczyg1:
> @miczyg1 shouldn't we use this https://github.com/Dasharo/coreboot/tree/common-base-rebased ? Please do cleanup of branches because its getting messy.

Ahh ok this PR is for old common-base... Allright

PR #213, opened 2022-07-11
soc/power9/pstates.c: handle disabled WOF properly

SergiiDmytruk:
There are two reasons it can be disabled:
 - chip doesn't support WOF
 - no WOF table for the chip was found


krystian-hebel (commented):
Two additional changes are required to flags in `fill_homer_for_chip()`:
- bit 3 in CME QM flags is set only if WOF enabled
- bit 3 in PGPE flags is set only if WOF enabled

https://github.com/open-power/hostboot/blob/master/src/import/chips/p9/procedures/hwp/pm/p9_hcode_image_build.C#L1308

PR #216, opened 2022-07-26
mb/protectli/vault_cml/bootblock.c: set WDT timeout to 0

mkopec:
Set the watchdog timer timeout to 0. The default value of 56 minutes caused unwanted reboots.

### DIFF
--- a/src/superio/ite/common/early_serial.c
+++ b/src/superio/ite/common/early_serial.c
@@ -7,12 +7,14 @@
 #include "ite.h"
 
 /* Global configuration registers. */
-#define ITE_CONFIG_REG_CC	0x02 /* Configure Control (write-only). */
-#define ITE_CONFIG_REG_LDN	0x07 /* Logical Device Number. */
-#define ITE_CONFIG_REG_CLOCKSEL	0x23 /* Clock Selection. */
-#define ITE_CONFIG_REG_SWSUSP	0x24 /* Software Suspend, Flash I/F. */
-#define ITE_CONFIG_REG_MFC	0x2a /* multi function pin */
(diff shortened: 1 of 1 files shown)


src/mainboard/protectli/vault_cml/bootblock.c
```diff
@@ -19,6 +19,8 @@ void bootblock_mainboard_early_init(void)
 	ite_reg_write(GPIO_DEV, 0x29, 0xc0);
 	/* External CLKIN PCICLK */
 	ite_reg_write(GPIO_DEV, 0x71, 0x08);
+	/* Set WDT timeout to 0 */
+	ite_reg_write(GPIO_DEV, 0x73, 0x00);
```
miczyg1, line 23: Let's add it to `src/superio/ite/common/early_serial.c: ite_kill_watchdog` function with the register definition.

miczyg1 (changes requested):

miczyg1:
@mkopec sending it upstream when migrated to common code would be great.

src/mainboard/protectli/vault_cml/bootblock.c
mkopec: Would have to check if it doesn't break other SuperIOs - although it should be fine since that's also how [the Linux it87_wdt driver](https://github.com/torvalds/linux/blob/e0dccc3b76fb35bb257b4118367a883073d7390e/drivers/watchdog/it87_wdt.c#L206) stops the watchdog

mkopec: Done, please check

macpijan:
We can continue review upstream if needed.

PR #218, opened 2022-08-03
Protectli vault cml: Disable default secure boot

Stojek139808:
Added parameter `SECURE_BOOT_DEFAULT_ENABLE=FALSE` and bumped the edk2 revision after rebasing it.


mkopec (approved):
Vboot changes should've been merged into `develop` instead of `release`, but thankfully this is not a problem here.

PR #225, opened 2022-08-22
configs: add no-TPM configs for KGPE-D16

krystian-hebel:
Signed-off-by: Krystian Hebel <krystian.hebel@3mdeb.com>


miczyg1:
@krystian-hebel where are the changes for VBOOT_NO_TPM Kconfig option? Solely removing TPM from configuration is not sufficient IIRC

krystian-hebel:
@miczyg1 based on https://github.com/Dasharo/coreboot/blob/asus_kgpe-d16/notpm_configs/src/security/vboot/Makefile.inc#L113 I thought this would suffice. `tpm/tspi/tspi.c` that [is changed with VBOOT_NO_TPM](https://github.com/Dasharo/coreboot/pull/191/files/c53b956eae601c2e27e8ac9047f5d46c71e1b17c..f056a9d7b857bead3ef5e4f75397208618010b72#diff-154fbdfb0b02f8a77b8f64a02671aaf370e8b5a17383722e9e852e78ccf3d9bdL101) isn't compiled without TPM enabled ([TPM1](https://github.com/Dasharo/coreboot/blob/f056a9d7b857bead3ef5e4f75397208618010b72/src/security/tpm/Makefile.inc#L5) or [TPM2](https://github.com/Dasharo/coreboot/blob/f056a9d7b857bead3ef5e4f75397208618010b72/src/security/tpm/Makefile.inc#L23)) and there are no link errors, so I assume this is enough. I guess this is something that would be best tested on hardware.

miczyg1:
@mkopec can you comment on this?

mkopec:
Correct, Vboot will work correctly if TPM is disabled in Kconfig. VBOOT_NO_TPM is for when you have VBOOT & TPM enabled but don't know if a TPM will always be present

miczyg1:
I see, thank @mkopec . Now quetion is if we should add 3 more targets instead of adding VBOOT_NO_TPM and keep the same number of targets? I am in favor of the latter

krystian-hebel:
Then to circumvent vboot you would have to just unplug the TPM, which is way easier and requires less tools than external flashing. This poses a security threat.

mkopec:
No, this would not disable vboot completely, Vboot still works without a TPM. The problem is that without `VBOOT_NO_TPM`, Vboot tries to initialize the TPM in bootblock, make measurements early, and access secdata (TPM NVRAM) for rollback protection. `VBOOT_NO_TPM` disables NVRAM accesses (`VBOOT_MOCK_SECDATA`) and causes TPM measurements to be logged into the TCPA log until the TPM is initialized in ramstage. Now that may be a reduction in security but Vboot still does all the verification it usually does.

miczyg1:
> Then to circumvent vboot you would have to just unplug the TPM, which is way easier and requires less tools than external flashing. This poses a security threat.

The only difference in my preferred approach is that you don't lose the measured boot capability in the firmware , all you have to do is to plug the TPM. With your approach, you would have to flash TPM enabled firmware.

krystian-hebel:
And why is that a problem? If user wants to use TPM, he/she probably wants to use it at each boot, and not change it dynamically, so flashing firmware once shouldn't be an issue.

miczyg1:
But it is an issue for us to maintain lots of build targets

miczyg1 (approved):

PR #227, opened 2022-09-01
Asus kgpe d16/rel v0.4.0

krystian-hebel:


miczyg1:
Apparently there are some conflicts. Also I wonder if we would need GH Actions jobs for the new build targets... But I am not sure if GH Actions still work... Also an update on dasharo/dasharo repo for YAMLs would bee needed as well.

krystian-hebel:
Conflicts are reported only by GUI, `git merge --ff-only` passes without issue. As for YAMLs, this is already prepared, but we probably shouldn't discuss it here :wink:

miczyg1 (approved):

PR #232, opened 2022-09-20
Audio for ADL-P

Stojek139808:


src/mainboard/clevo/adl-p/variants/ns70pu/gpio.c
```diff
@@ -3,6 +3,13 @@
 #include <mainboard/gpio.h>
 #include <soc/gpio.h>
 
+#ifndef PAD_CFG_GPIO_BIDIRECT
+#define PAD_CFG_GPIO_BIDIRECT(pad, val, pull, rst, trig, own)		\
```
macpijan, line 7: @Stojek139808 What is the source of this macro? Is it really specific to to the NS70PU model to go there?

Stojek139808: It's copied from other gpio headers. On this branch I found it here:
```
src/mainboard/51nb/x210/gpio.h
src/mainboard/razer/blade_stealth_kbl/gpio.h
```
But now that you mention this, it's probably better to move this from `gpio.c` to `gpio.h`.

mkopec: Could move this to common code, there's a TODO for this in every instance this `#define` appears. Can push it upstream afterwards, too

PR #233, opened 2022-09-21
UEFI boot

Stojek139808:


configs/config.clevo_ns70pu
```diff
@@ -9,6 +9,8 @@ CONFIG_CBFS_SIZE=0xA00000
 CONFIG_MAINBOARD_SMBIOS_PRODUCT_NAME="NS5x_NS7xPU"
 # CONFIG_USE_PM_ACPI_TIMER is not set
 CONFIG_CONSOLE_CBMEM_BUFFER_SIZE=0x20000
+CONFIG_EDK2_BOOT_TIMEOUT=4
+CONFIG_USE_LEGACY_8254_TIMER=y
```
mkopec, line 13: Unneeded, causes sleep issues and increased power draw. Please remove

Stojek139808: Removed.

configs/config.clevo_ns70pu
```diff
@@ -9,6 +9,8 @@ CONFIG_CBFS_SIZE=0xA00000
 CONFIG_MAINBOARD_SMBIOS_PRODUCT_NAME="NS5x_NS7xPU"
 # CONFIG_USE_PM_ACPI_TIMER is not set
 CONFIG_CONSOLE_CBMEM_BUFFER_SIZE=0x20000
+CONFIG_EDK2_BOOT_TIMEOUT=4
```
mkopec, line 12: Shouldn't this be 2 on NovaCustom laptops?

Stojek139808: Doesn't change that much but changed it to 2, now it boots a bit faster since it waits 2s less.

PR #237, opened 2022-09-27
src/mainboard/novacustom: add NovaCustom as OEM for clevo laptops

mkopec:
Modified slightly from the previous OEM structure to accomodate multiple generations of devices.


macpijan:
If we will target such MRs to `release` branch directly we will keep loosing stuff on `develop`

macpijan (approved):

PR #239, opened 2022-10-06
configs/config.clevo_ns70pu: include IFD and ME binaries

miczyg1:


mkopec:
@macpijan Can we redistribute these now? Previously I added ME and IFD binaries separately after building

miczyg1:
@mkopec we can include a separate config for IFD/ME-less build. We won't distribute these

miczyg1:
Added separate config for full build

PR #253, opened 2022-10-28
msi_ms7d25: BIOS Lock and SMM_BWP, enable EDK2 setup password menu

mkopec:
Enable EDK2 setup password menu. Detect BIOS lock and SMM_BWP options from EFI variables.


mkopec:
I'm not too happy with including intelpch/lockdown into SMM, but I wasn't sure where else to put `is_smm_bwp_permitted()`

payloads/external/edk2/Kconfig
```diff
@@ -268,6 +268,12 @@ config EDK2_OPAL_PASSWORD
 	help
 	  Enables TCG OPAL disk password in the TianoCore Payload.
 
+config EDK2_SETUP_PASSWORD
+	bool "Enable TianoCore setup menu password"
+	default n
+	help
+	  Enables setup menu password in the TianoCore Payload.
```
miczyg1, line 275: I would make it default to enabled. This is not something we want t configure I guess. So no need to specify Kconfig for it

miczyg1 (changes requested):

payloads/external/edk2/Kconfig
mkopec: Done. Removed the option entirely

miczyg1 (approved):

PR #254, opened 2022-10-28
Watchdog support

miczyg1:


mkopec:
If we initialize the watchdog in bootblock, a user updating from v1.0.0 [flashing only the RW sections](https://docs.dasharo.com/variants/msi_z690/firmware-update/#version-v100-or-newer) won't be able to utilize this. I'd rather put in in romstage or ramstage instead

miczyg1:
> If we initialize the watchdog in bootblock, a user updating from v1.0.0 [flashing only the RW sections](https://docs.dasharo.com/variants/msi_z690/firmware-update/#version-v100-or-newer) won't be able to utilize this. I'd rather put in in romstage or ramstage instead

Yes, but then FSP would detect unexpected reset if vboot would want to reboot to recovery for example. There is no good solution to this. By the way, we do a major update and bump minor version of the firmware as well. In situations like this we may advise to flash recovery section as well. For patch version updates, we will avoid touching recovery region @mkopec .

mkopec:
Understood.

mkopec (approved):

PR #255, opened 2022-10-29
Msi ms7d25/me disable

mkopec:


src/soc/intel/common/block/cse/cse.c
```diff
@@ -1071,6 +1076,85 @@ void cse_control_global_reset_lock(void)
 		pmc_global_reset_enable(false);
 }
 
+static const EFI_GUID dasharo_system_features_guid = {
```
miczyg1, line 1079: This also have to go to preprocessor IF because if not used compiler will give warning. Maybe put it inside cse_get_me_disable_mode as there is not other use of this guide elsewhere in the file

miczyg1 (changes requested):

src/soc/intel/common/block/cse/cse.c
mkopec: Done

miczyg1 (approved):

PR #260, opened 2022-11-09
payloads/external/edk2: Add MTRR programming option

miczyg1:


payloads/external/edk2/Makefile
```diff
@@ -121,10 +121,14 @@ endif
 ifeq ($(CONFIG_EDK2_TPM_ENABLE),y)
 BUILD_STR += -D TPM_ENABLE=TRUE
 endif
+# NETWORK_LAN_ROM		= FALSE
```
macpijan, line 124: Commit says only about disabling MTRR. Do we also need that line here?

miczyg1: Forgot to add the comment with the default value in previous commits/PRs for this option. Good to have but not needed.

macpijan: Maybe it should be a separate commit then, if note related to the MTTR option? Would make it easier when rebasing and cleaning up patches.

miczyg1: Made it a separate commit

macpijan: Thanks

PR #282, opened 2023-01-05
Novacustom ACPI thermal zones

mkopec:


src/ec/system76/ec/acpi/thermal.asl
```diff
@@ -0,0 +1,162 @@
...
+
+		Name (TFST, Package ()
+		{
+			0,	// Revision
+			0x00,	// Control
+			0x00	// Speed
+		})
+
+		/* _FST: Fan current Status */
+		Method (_FST, 0, Serialized,,PkgObj)
```
miczyg1, line 26: I think only fan status is needed here. All Fan methods are optional according to ACPI spec.

miczyg1 (changes requested):

src/ec/system76/ec/acpi/thermal.asl
mkopec: ACPI spec confusingly also [states](https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/11_Thermal_Management/fan-device.html): `While the Fan Device and its associated objects are optional, if the Fan Device is implemented by the platform, all objects listed in the table above are required and must be provided.`

miczyg1: Ahh okay...

miczyg1 (approved):

src/ec/system76/ec/acpi/thermal.asl
mkopec: And linux also expects all 4 to be implemented: https://github.com/torvalds/linux/blob/7dd4b804e08041ff56c88bdd8da742d14b17ed25/drivers/acpi/fan_core.c#L206-L212

PR #283, opened 2023-01-10
src/mainboard/clevo/adl-p: S3 suspend support

miczyg1:
Tested on NS70PU, both Ubuntu 22.04 and Windows 11.
NV40PZ needs  testing.

Needs further testing for reliability (fwts 30 cycles for example)


miczyg1:
Tested on NS50PU Ubuntu 22.04 and works.
Tested on NV41PZ Fedora 37 and works.

mkopec:
Is it necessary to remove RTD3 for everyting? Does that not impact power usage at idle?

miczyg1:
@mkopec I had to remove it because otherwise all RTD3 devices are unresponsive and Linux hangs (cannot talk to NVMe for example). I don't know why OS did not call _ON methods , etc. to take it out from D3cold. Maybe something else failed, I don't know, because I have no means to debug what is going on in the Linux kernel.

mkopec (approved):
Thanks

XGIVolari:
Sorry for the noise but as an owner of an affected device, (NovaCustom NS51 ADL-P) I'd like to raise my concerns about this decision. AFAICT, this PR disables not only RTD3, but s0ix entirely. I really don't think either of these would be worthwhile tradeoff only to enable S3, and especially not both at once. 
Given that most modern laptops don't support S3 anyways and are designed to work with s0ix, I doubt whether S3 is even possible at all without serious tradeoffs such as this. Personally, I do not expect any modern laptop to support S3 and I am much more concerned about correctly functioning s0ix. Thank you for your consideration.

miczyg1:
@XGIVolari while I understand your concerns, the switch to S3 is dictated by this issue: https://github.com/Dasharo/dasharo-issues/issues/266

tl;dr NovaCustom laptops are going to be Qubes OS certified, but for that to happen, S3 must be working. S0ix is not supported by Xen+Linux, so S3 is the only option. Otherwise, we would keep S0ix.

PS: Also S0ix has also some security concerns, as the machine is still operable to some extent while sleeping. With S3 the machine is in halted state until the waking vector is applied.

XGIVolari:
I see. Thanks for your insightful response.

wessel-novacustom:
Hi @XGIVolari, thank you for sharing your concern and welcome to the community.

There are more concerns about S0ix modern standby than S3 suspend mode. Both have down-sides, but we expect S3 to have a better battery life while in sleep (S0ix has a faster startup time from sleep, but that doesn't seem to be a big deal to our users).

The main decision is because of Qubes OS that we want to support. Currently, they don't support modern standby (S0ix).

We hope that S0ix will improve in the future so that we can switch back to S0ix again.

XGIVolari:
Would it be feasible to implement a toggle between s0ix/S3 inside the firmware setup menu? That way, we could get the best of both worlds.

miczyg1:
@XGIVolari I would have to look closer at it, but maybe.

pietrushnic:
@XGIVolari we appreciate your feedback and we will try whatever is reasonable to satisfy your needs.

@miczyg1 let's cross-link issue related to S0ix in Qubes OS, just for future reference. I believe S0ix vs S3 topic may be of interest for wider community that's why, if there are any security, reliability or performance considerations those should be clearly expressed. We would like to avoid situation when whenever (until reliable S0ix implementation in Xen and Linux) question about both arise users would have to search through issues. This is definitely topic for NovaCustom Dasharo FAQ. We don't have to dive in guts of the problem, we just have to provide information about decision process and pointers where to find more about the topic, discuss or contribute.

PR #284, opened 2023-01-16
Persistent keyboard backlight

mkopec:
Tested on NS50PU. Works, but setting is overriden later in the boot process by the `system76_acpi` driver. Should probably stop relying on it and either use standard ACPI interfaces or have our own OS driver.

Depends on https://github.com/Dasharo/ec/pull/17


mkopec:
Added w/a for os driver overriding our settings. Tested on tgl model with white backlit keyboard, everything works fine. Added unrelated fix for nv4x_tgl compilation.

miczyg1 (approved):

PR #285, opened 2023-01-23
Clevo / Runtime configurable fan curve

mkopec:


mkopec:
@miczyg1 Ready for review now, just will need to update miniconfigs when https://github.com/Dasharo/DasharoModulePkg/pull/13 gets merged

src/mainboard/clevo/tgl-u/ramstage.c
```diff
@@ -31,13 +31,30 @@ struct smfi_cmd_set_fan_curve {
 	struct fan_curve curve;
 };
 
-#define FAN_CURVE_DEFAULT {     \
+#define FAN_CURVE_SILENT {      \
   { .temp = 0,   .duty = 25  }, \
```
miczyg1, line 35: coreboot uses TAB indents (8 character long))

src/mainboard/clevo/tgl-u/ramstage.c
```diff
@@ -31,13 +31,30 @@ struct smfi_cmd_set_fan_curve {
...
   { .temp = 65,  .duty = 30  }, \
   { .temp = 75,  .duty = 35  }, \
   { .temp = 100, .duty = 100 }  \
 }
 
+#define FAN_CURVE_PERFORMANCE { \
+  { .temp = 0,   .duty = 25 },  \
```
miczyg1, line 42: as above

src/mainboard/clevo/tgl-u/ramstage.c
```diff
@@ -31,13 +31,30 @@ struct smfi_cmd_set_fan_curve {
...
+  { .temp = 100, .duty = 100}   \
+}
+
+#define FAN_CURVE_OPTION_SILENT 0
+#define FAN_CURVE_OPTION_PERFORMANCE 1
+
+#define FAN_CURVE_OPTION_DEFAULT FAN_CURVE_OPTION_SILENT
+
+static struct fan_curve fan_curves[] = {
+  [FAN_CURVE_OPTION_SILENT]      = FAN_CURVE_SILENT,
```
miczyg1, line 54: as above

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -29,13 +29,31 @@ struct smfi_cmd_set_fan_curve {
...
-#define FAN_CURVE_DEFAULT {     \
+#define FAN_CURVE_SILENT {      \
   { .temp = 0,   .duty = 25  }, \
   { .temp = 65,  .duty = 30  }, \
   { .temp = 75,  .duty = 35  }, \
   { .temp = 100, .duty = 100 }  \
 }
 
+#define FAN_CURVE_PERFORMANCE { \
+  { .temp = 0,   .duty = 25 },  \
```
miczyg1, line 40: as above

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -29,13 +29,31 @@ struct smfi_cmd_set_fan_curve {
 	struct fan_curve curve;
 };
 
-#define FAN_CURVE_DEFAULT {     \
+#define FAN_CURVE_SILENT {      \
   { .temp = 0,   .duty = 25  }, \
```
miczyg1, line 33: as above

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -29,13 +29,31 @@ struct smfi_cmd_set_fan_curve {
...
+  { .temp = 100, .duty = 100}   \
+}
+
+#define FAN_CURVE_OPTION_SILENT 0
+#define FAN_CURVE_OPTION_PERFORMANCE 1
+
+#define FAN_CURVE_OPTION_DEFAULT FAN_CURVE_OPTION_SILENT
+
+static struct fan_curve fan_curves[] = {
+  [FAN_CURVE_OPTION_SILENT]      = FAN_CURVE_SILENT,
```
miczyg1, line 52: as above

miczyg1 (changes requested):

src/mainboard/clevo/tgl-u/ramstage.c
mkopec: ofc, fixed, copypaste error when moving this code from DasharoModulePkg

src/mainboard/clevo/tgl-u/ramstage.c
mkopec: fixed

src/mainboard/clevo/adl-p/ramstage.c
mkopec: fixed

src/mainboard/clevo/adl-p/ramstage.c
mkopec: fixed

src/mainboard/clevo/adl-p/ramstage.c
mkopec: fixed

src/mainboard/clevo/tgl-u/ramstage.c
mkopec: done

mkopec:
Wired up the new fan curve PCD and organized the DasharoModulePkg-specific Kconfig options somewhat, so menuconfig displays them in a hierarchy.

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -3,6 +3,56 @@
...
+
+#define FAN_CURVE_OPTION_SILENT 0
+#define FAN_CURVE_OPTION_PERFORMANCE 1
+
+#define FAN_CURVE_OPTION_DEFAULT FAN_CURVE_OPTION_SILENT
+
+static struct fan_curve fan_curves[] = {
+	[FAN_CURVE_OPTION_SILENT]      = FAN_CURVE_SILENT,
+	[FAN_CURVE_OPTION_PERFORMANCE] = FAN_CURVE_PERFORMANCE
+};
```
miczyg1, line 54: I think it would be better to have static structs for that instead of arrays and macros:

```
struct fan_curve fan_curve_silent = {
	{ .temp = 0,   .duty = 25  },
	{ .temp = 65,  .duty = 30  },
	{ .temp = 75,  .duty = 35  },
	{ .temp = 100, .duty = 100 } 
}
```

Etc. And then:

```
switch (selection) {
case FAN_CURVE_OPTION_SILENT: cmd.curve = fan_curve_silent; break;
case FAN_CURVE_OPTION_PERFORMANCE: cmd.curve = fan_curve_performance; break;
default: cmd.curve = fan_curve_silent; break;
}
```

miczyg1: Also the structs should have `__packed` attribute right? Because the yare used as raw input to EC CMD

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -3,6 +3,56 @@
...
 #include <soc/ramstage.h>
 #include <smbios.h>
+#include <ec/system76/ec/commands.h>
+#include <drivers/efi/efivars.h>
+#include <fmap.h>
+#include <smmstore.h>
+
+#include <Uefi/UefiBaseType.h>
+
+#include <lib.h>
```
miczyg1, line 13: Please order the includes in alphabetical order.  `Uefi/UefiBaseType.h` as a foreign include may be separated as is now.

miczyg1 (changes requested):

src/mainboard/clevo/adl-p/ramstage.c
mkopec: Okay, done as per suggestion. And yes the structs clearly needed to be packed

src/mainboard/clevo/adl-p/ramstage.c
mkopec: Alphabetized

payloads/external/edk2/Makefile
```diff
@@ -153,6 +153,13 @@ endif
 ifeq ($(CONFIG_EDK2_DASHARO_CHIPSET_CONFIG),y)
 BUILD_STR += --pcd gDasharoSystemFeaturesTokenSpaceGuid.PcdShowChipsetMenu=TRUE
 endif
+# DASHARO_POWER_CONFIG		= FALSE
```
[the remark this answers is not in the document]
mkopec: hmm, i see this notation used on different branches as well. Is there an example of what this should look like instead?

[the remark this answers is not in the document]
miczyg1:
```suggestion
```

Just like that if we use PCDs. A default value for this PCD should be FALSE and that probably applies to all submenus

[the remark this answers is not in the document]
mkopec: okay, done

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -65,9 +108,54 @@ void mainboard_silicon_init_params(FSP_S_CONFIG *params)
...
+	size = sizeof(selection);
+	ret = efi_fv_get_option(&rdev, &dasharo_system_features_guid, "FanCurveOption", &selection, &size);
+	if (ret != CB_SUCCESS)
+		printk(BIOS_DEBUG, "fan: failed to read curve selection from EFI vars, using board default\n");
+efi_err:
+#endif
+
+	switch (selection) {
+	case FAN_CURVE_OPTION_SILENT:
+		cmd.curve = fan_curve_silent;
```
miczyg1, line 138: does such assignment work as expected? For simple structures it probably would, but to be sure memcpy will be safest. @krystian-hebel can you confirm?

krystian-hebel: In this case it doesn't make a difference, basic assignment is safe to use. `memcpy` helps in two cases:
- an incomplete type for which size of copied data isn't available, e.g. when structure ends with unbounded array (like `uint8_t buf[];`) or when more than one element of array is to be copied, or when compiler doesn't have full definition of type,
- struct has unaligned fields on architectures that can't access such data directly (but this usually is mitigated by compiler anyway).

Neither of those would help if the structure has pointers to data that would also have to be copied.

miczyg1 (approved):

PR #287, opened 2023-01-25
EC branding

Stojek139808:


src/ec/system76/ec/Kconfig
```diff
@@ -22,3 +22,11 @@ config EC_SYSTEM76_EC_OLED
 	depends on EC_SYSTEM76_EC
 	bool
 	default n
+
+config EC_SYSTEM76_EC_ACPI_DEVICE_HID
+	string
+	default "17761776"
+
+config EC_SYSTEM76_EC_ACPI_DEVICE_NAME
```
mkopec, line 30: Kconfig won't allow us to have a bare string without `"`s here, and iASL won't allow us to pass a string type as device name. Let's not add this option.

src/ec/system76/ec/acpi/s76.asl
```diff
@@ -8,13 +8,13 @@
...
 Device (S76D) {
-	Name (_HID, "17761776")
+	Name (_HID, CONFIG_EC_SYSTEM76_EC_ACPI_DEVICE_HID)
 	Name (_UID, 0)
 	// Hide the device so that Windows does not warn about a missing driver.
 	Name (_STA, 0xB)
 
 	Method (RSET, 0, Serialized) {
-		Debug = "S76D: RSET"
+		Debug = ": RSET"
```
mkopec, line 17:
```suggestion
		Debug = "S76D: RSET"
```

Let's put it the way it was before.

src/ec/system76/ec/Kconfig
```diff
@@ -22,3 +22,11 @@ config EC_SYSTEM76_EC_OLED
...
+	string
+	default S76D
```
mkopec, line 32:
```suggestion
```

src/ec/system76/ec/Kconfig
mkopec: See suggested change below.

mkopec (approved):

PR #295, opened 2023-02-13
S3 for NV40MZ

mkopec:
Same as it was done for NS5x TGL-U.


mkopec:
Updated finalize.c to match ADL implementation (implemented PMC comm natively in coreboot instead of calling FspNotify). Disabled all FSP notifications and now resume no longer hangs in FspNotify(PciEnumDone) anymore. I still get `IPC_STS.error_code 0x4` in logs but it seems to be harmless. NV40MZ completes 20 suspend/resume cycles without issue.

mkopec:
And now after cleaning up the patches it's broken again... Needs more work.

mkopec:
Logs suggest that coreboot jumps to the OS waking vector but nothing happens afterwards. Need to debug with a device with a debug interface, so I'm marking this as draft until we can debug this properly. Pushed some debug leftovers, too.

miczyg1:
@mkopec is it still valid PR? I think we will go with S3/S0ix switch right?

mkopec:
@miczyg1 We'll probably go with the switch, yes. We can close this but this branch still contains useful code like the native PMC lockdown in coreboot

miczyg1:
@mkopec let's clean it up then (when you get some free cycles) and we can merge it to develop instead of closing

mkopec:
Merged in https://github.com/Dasharo/coreboot/pull/419, closing.

PR #302, opened 2023-02-24
ec/system76/ec/acpi/s76.asl: reenable keyboard on reboot

mkopec:
When rebooting, the lid switch state does not change, so the EC will not automatically reenable backlight. Call EKBL during boot to fix this.


mkopec:
Fixes https://github.com/Dasharo/dasharo-issues/issues/349

miczyg1 (approved):

PR #306, opened 2023-03-03
MSI MS-7D25 ACPI cleanup

miczyg1:


miczyg1:
Also cleaning up ACPI errors about missing devices.

PR #311, opened 2023-03-08
Sleep type option

miczyg1:


mkopec (changes requested):
Like I commented in https://github.com/Dasharo/ec/pull/30#pullrequestreview-1334475968, please drop the EC related changes, they're not needed and break opportunistic S0ix.

miczyg1:
> Like I commented in [Dasharo/ec#30 (review)](https://github.com/Dasharo/ec/pull/30#pullrequestreview-1334475968), please drop the EC related changes, they're not needed and break opportunistic S0ix.
> 
> Other than that, the option works correctly.

Removed the EC command and SMFI call

mkopec (approved):

PR #312, opened 2023-03-09
src/mainboard/clevo/adl-p/Kconfig: unselect HAVE_ACPI_RESUME

mkopec:


mkopec:
right, in that case this can be closed.

PR #322, opened 2023-03-16
OS issues

Stojek139808:


miczyg1:
1. This isn't probably a fix.
2. Furthermore we don't use S0ix so S0ix should be disabled by the board.
3. I have switched the CPU init method to use FSP, there are probably more problems becasue of that (C states, frequency, etc.)

miczyg1 (approved):
I guess we don't have any other way. AMI BIOS does not expose any C states in ACPI too

PR #323, opened 2023-03-16
Clevo/nv4x adl v1.6.0 fixes

mkopec:
- edk2: add edk2 logging to s76 ec (depends on https://github.com/Dasharo/edk2/pull/54)
- smmstore_init: retry calling SMI handler up to 5 times upon failure


src/mainboard/clevo/adl-p/devicetree.cb
```diff
@@ -20,6 +20,13 @@ chip soc/intel/alderlake
 	# Thermal
 	register "tcc_offset" = "8"
 
+	# Acoustic settings
+	register "acoustic_noise_mitigation" = "1"
+	register "slow_slew_rate[VR_DOMAIN_IA]" = "SLEW_FAST_8"
+	register "slow_slew_rate[VR_DOMAIN_GT]" = "SLEW_FAST_8"
+	register "fast_pkg_c_ramp_disable[VR_DOMAIN_IA]" = "1"
+	register "fast_pkg_c_ramp_disable[VR_DOMAIN_GT]" = "1"
```
miczyg1, line 28: Is the noise reduced with these settings? How did you determine the settings?

src/drivers/smmstore/ramstage.c
```diff
@@ -57,14 +59,24 @@ static void init_store(void *unused)
...
+			: "=a" (eax)
+			: "a" ((SMMSTORE_CMD_INIT << 8) | APM_CNT_SMMSTORE),
+			  "b" (ebx),
+			  "d" (APM_CNT)
+			: "memory");
+
+		if (eax == SMMSTORE_RET_SUCCESS)
+			break;
+
+		mdelay(1);
```
miczyg1, line 78: @krystian-hebel not sure if retries are good here.

@mkopec have you looked at SMMSTORE_CMD_INIT code? Is it reentrant?

mkopec: This is based on the comment and code in edk2's SMMSTORE v2 driver: https://github.com/MrChromebox/edk2/commit/27854bc8c5e7147f01e50404f6e5b16ffedf08b8#diff-3ce60dbfce1b2c4fb8b00b51d566b6a056a567c819ebedbd4e74fcb0834cdcc9R28-R41

If a previous SMMSTORE_CMD_INIT proceeds successfully but the value of eax isn't successful then the next execution will return SMMSTORE_RET_FAILURE - but assuming the comment in the link above is correct, that will never happen: `[...] in which case the Smi handler will do nothing as only synchronous Smis are handled`

krystian-hebel: https://github.com/Dasharo/coreboot/blob/0c6a6f7d7ac426e0eee5d6ad26846bb68295c6ef/src/drivers/smmstore/store.c#L303 should be safe to call, but would return error anyway. It would be nice to have a way of returning "already initialized" status.

src/mainboard/clevo/adl-p/devicetree.cb
mkopec: Removed in the latest patchset.

krystian-hebel (approved):

PR #330, opened 2023-03-21
Ec version strings for SOL

miczyg1:


mkopec (dismissed):

miczyg1:
@macpijan I see develop branch is undergoing major changes due to enterprise builds so I was not sure if I should set the target as develop. So I used release as more "stable" target. Question is when it should be incorporated.

macpijan:
> I see develop branch is undergoing major changes due to enterprise builds

That is why my idea and suggestion was **not** to include these changes into develop, but to https://github.com/Dasharo/coreboot/tree/clevo/dasharo-enteprise as I did in: https://github.com/Dasharo/coreboot/pull/320

macpijan:
@miczyg1 

> I see develop branch is undergoing major changes due to enterprise builds so

What major changes in particular?

I can see that the DMA protection went to develop, but the other features target the integration branch: https://github.com/Dasharo/coreboot/tree/clevo/dasharo-enteprise

I think that it makes sense that this SoL goes to develop, as each feature.

mkopec (approved):

PR #332, opened 2023-03-27
Config file correction

3mdeb-karolzmyslowski:
For the following platform:
  - Novacustom NV4 Tiger Lake
  - Novacustom NS5 Tiger Lake
there is need to synchronize option, to add functionalities:
  - security boot menu options
  - network boot enable

Altered config, to meet requirements, against mentioned platform, synchronize with Novacustom NV4/NS5 Alder Lake platform.


miczyg1:
These options should live on the dasharo-enterprise branch, no need to synchronize I think. Closing

PR #356, opened 2023-07-03
Raptor cs talos 2/cbmem use sysfs

krystian-hebel:


SergiiDmytruk (approved):
LGTM. Running `cbmem -t` on Talos produces
```
352321536 entries total

   0:1st timestamp                                     4,495,843,021,643,841,536 (0)
...
```
With the last line repeated, but I don't think it's related to these changes.  Might be wrong endianness.

PR #363, opened 2023-07-21
Msi ms7d25 rel v1.1.2 ms7e06 rel v0.9.0

miczyg1:


.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
+name: MSI PRO Z-690A
```
mkopec, line 1: CI probably needs updating to build for both platforms

miczyg1: Yup, I am still working on polishing such things

miczyg1: Added a bit later: https://github.com/Dasharo/coreboot/commit/9899b427131f4fe081918cbe779fab72407527c5

PR #364, opened 2023-08-01
Novacustom power limits

mkopec:
Set default limits as if the barrel jack AC adapter is used. These limits are adjusted later by EC.

### DIFF
--- a/src/mainboard/clevo/adl-p/variants/ns50pu/overridetree.cb
+++ b/src/mainboard/clevo/adl-p/variants/ns50pu/overridetree.cb
@@ -1,10 +1,14 @@
 chip soc/intel/alderlake
-	# HACK: Limit PL4 to PL2 to prevent power-off when system is booted on
-	# battery power. This seems to only happen with the i7 units.
+	# Default values at boot, to be adjusted by EC at runtime according
+    # to available wall brick capabilities.
 	register "power_limits_config[ADL_P_282_482_28W_CORE]" = "{
 		.tdp_pl1_override = 20,
 		.tdp_pl2_override = 56,
-		.tdp_pl4 = 56, // FIXME: Set to 65
+		.tdp_psyspl2 = 90,
+		.tdp_psyspl3 = 92,
+		.tdp_psyspl3_time = 6,
+		.tdp_psyspl3_dutycycle = 4,
+		.tdp_pl4 = 65,
 	}"
 
 	# GPE configuration
--- a/src/mainboard/clevo/adl-p/variants/nv40pz/overridetree.cb
+++ b/src/mainboard/clevo/adl-p/variants/nv40pz/overridetree.cb
@@ -1,8 +1,14 @@
 chip soc/intel/alderlake
+	# Default values at boot, to be adjusted by EC at runtime according
+    # to available wall brick capabilities.
 	register "power_limits_config[ADL_P_282_482_28W_CORE]" = "{
 		.tdp_pl1_override = 28,
 		.tdp_pl2_override = 60,
-		.tdp_pl4 = 90,
+		.tdp_psyspl2 = 90,
+		.tdp_psyspl3 = 92,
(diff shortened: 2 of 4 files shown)


src/mainboard/clevo/adl-p/variants/ns50pu/overridetree.cb
```diff
@@ -1,10 +1,14 @@
 chip soc/intel/alderlake
-	# HACK: Limit PL4 to PL2 to prevent power-off when system is booted on
-	# battery power. This seems to only happen with the i7 units.
+	# Default values at boot, to be adjusted by EC at runtime according
+    # to available wall brick capabilities.
 	register "power_limits_config[ADL_P_282_482_28W_CORE]" = "{
 		.tdp_pl1_override = 20,
 		.tdp_pl2_override = 56,
-		.tdp_pl4 = 56, // FIXME: Set to 65
+		.tdp_psyspl2 = 90,
```
miczyg1, line 7: psysPL2 should be equal to AC brick ratio
psysPL3 should be psysPL2 + 2W per Intel recommendations

miczyg1 (changes requested):

src/mainboard/clevo/adl-p/variants/nv40pz/overridetree.cb
```diff
@@ -1,7 +1,13 @@
 chip soc/intel/alderlake
+	# Default values at boot, to be adjusted by EC at runtime according
+    # to available wall brick capabilities.
 	register "power_limits_config[ADL_P_282_482_28W_CORE]" = "{
 		.tdp_pl1_override = 28,
 		.tdp_pl2_override = 60,
```
miczyg1, line 6: BTW why NV4x has higher power limits? Does it have better cooling than NS5x?

src/mainboard/clevo/tgl-u/variants/ns50mu/devicetree.cb
```diff
@@ -1,4 +1,21 @@
 chip soc/intel/tigerlake
+# CPU (soc/intel/tigerlake/cpu.c)
+    register "power_limits_config[POWER_LIMITS_U_4_CORE]" = "{
+		.tdp_psyspl2 = 65,
```
miczyg1, line 4: psysPL1 and psysPL2 work concurrently with PL1 and PL2, we should probably set those as well.

BTW why we do not set psysPL1?

src/mainboard/clevo/adl-p/variants/nv40pz/overridetree.cb
mkopec: Correct, it has dual fans, NS5x has one fan only.

src/mainboard/clevo/tgl-u/variants/ns50mu/devicetree.cb
mkopec: PL2 / 3 are programmed to base TDP and base TDP * 125 in by coreboot by default

src/mainboard/clevo/adl-p/variants/ns50pu/overridetree.cb
mkopec: done

src/mainboard/clevo/tgl-u/variants/ns50mu/devicetree.cb
mkopec: I'm not sure if we need PsysPL1, and what value would make sense here. Intel docs recommend setting it to device cooling capacity.

miczyg1: > PL2 / 3 are programmed to base TDP and base TDP * 125 by coreboot by default

Ahh right. 

> I don't believe the RoP power draw contributes significantly to thermals here.

Fine.

PR #370, opened 2023-08-14
Msi ms7d25/fum

krystian-hebel:


src/soc/intel/common/block/cse/cse.c
```diff
@@ -1095,6 +1095,12 @@ unsigned int cse_get_me_disable_mode(void)
 	if (smmstore_lookup_region(&rdev))
 		return 0;
 
+    /* Disable ME if in Firmware Update Mode */
+	size = sizeof(var);
+	ret = efi_fv_get_option(&rdev, &dasharo_system_features_guid, "FirmwareUpdateMode", &var, &size);
```
miczyg1, line 1100: Like I mentioned in other channel, it may be better to simply invoke `cse_hmrfpo_enable` on FUM path as it will ensure the descriptor is also unlocked, just in case. It will also cover cases where ME region is RO in the descriptor.

In such case we would also need to add `cse_hmrfpo_enable` implementation.

miczyg1, line 1100: Like I mentioned in other channel, it may be better to simply invoke `cse_hmrfpo_enable` on FUM path as it will ensure the descriptor is also unlocked, just in case. It will also cover cases where ME region is RO in the descriptor.

In such case we would also need to add `cse_hmrfpo_enable` implementation.

miczyg1 (changes requested):

src/soc/intel/common/block/cse/cse.c
mkopec: The ME_WE pin on the EC may be used for the same purpose, in a later iteration perhaps.

mkopec (dismissed):

src/soc/intel/common/block/cse/cse.c
miczyg1: https://github.com/Dasharo/dasharo-issues/issues/481

miczyg1 (approved):
Waiting for CI to pass (rebased on recent develop)

PR #377, opened 2023-08-24
Clevo/qcnfa222 wifi test

miczyg1:


miczyg1:
Found a few incorrect things in the code:
1. Some RTD3 GPIOs had a low state, despite all should have a high state to deactivate reset and activate power. Also the reset configs of RTD3 pads was not consistent.
2. NV4x ADL missed an enable GPIO for GLAN (see NV4XPZ service manual)
3. Added missing FSP UPDs for PCIe RP10 which is used in NV4x.

mkopec (approved):

PR #378, opened 2023-08-25
Wifi bt switch ns5x adl v1.6.0

Mixss:
I don't which branch should be a base branch


miczyg1:
> I don't which branch should be a base branch

Typically it is the one you have started from

miczyg1 (changes requested):

PR #383, opened 2023-08-30
[MSI] Enable PCIe resizeable BARs and control it via EFI variable

SergiiDmytruk:
Other changes:
 - drop `-j$(nproc)` from `build.sh` as it tends to break the build
 - fix one `Kconfig` which makes coreboot's pre-commit hook pass, it looks like a typo


build.sh
```diff
@@ -65,7 +65,7 @@ function build_image_z690 {
 	docker run --rm -t -u $UID -v $PWD:/home/coreboot/coreboot \
 		-v $HOME/.ssh:/home/coreboot/.ssh \
 		-w /home/coreboot/coreboot coreboot/coreboot-sdk:2021-09-23_b0d87f753c \
-		/bin/bash -c "make olddefconfig && make -j$(nproc)"
```
miczyg1, line 68: In what way the build breaks? @SergiiDmytruk 

I would like to keep the multiple jobs, otherwise the CI may make even longer to build (not even mentioning users)

SergiiDmytruk: There was some compilation error, restarting the build helped as it doesn't happen every time.  `make` also hanged after doing some changes and recompiling.

If it works more often than not, I'll just keep the change locally and will take a closer look if the failure repeats, I just thought `-j` was added by accident.

miczyg1: I have never experienced such issues TBH. Whenever I repeat the make after compilation issues it always builds with the fresh local changes.

miczyg1 (approved):

PR #387, opened 2023-09-04
src/soc/intel/alderlake/acpi.c: Fix reported C-states on desktop processors

miczyg1:
Desktop processors do not report C7 substate support in CPUID. Use C8 which
is the first supported lower state than C7 state, which is C8. This fixes
the unsupported C-state Firmware Bug in Linux dmesg.


miczyg1:
Checked turbostat and now the P-cores are able to enter the C7 state (C7 reported by the turbostat). Previously the cores were only able to enter C6.

mkopec:
CI seems to be failing on submodule clone

mkopec (approved):

miczyg1:
> CI seems to be failing on submodule clone

Yeah, restarting over and over and can't pass it

miczyg1:
CI passed, merging

PR #388, opened 2023-09-04
mb/msi/{ms7d25,ms7e06}: Reset audio codec

miczyg1:


miczyg1:
I did a few reboot cycles between Windows and Ubuntu using front and rear jacks wit ha headset. This time the playback always worked in both Windows and Linux

mkopec (approved):

PR #390, opened 2023-09-11
S0ix fixes after EC rebase

miczyg1:
S0ix PMC hack relies on reading ECOS field, so it has to be read no matter which sleep type we are entering

### DIFF
--- a/src/ec/system76/ec/acpi/ec.asl
+++ b/src/ec/system76/ec/acpi/ec.asl
@@ -9,6 +9,10 @@ Scope (\_SB) {
 	#include "s76.asl"
 }
 
+Scope (\_SB.PCI0.LPCB.RTC) {
+	#include "cmos.asl"
+}
+
 Device (\_SB.PCI0.LPCB.EC0)
 {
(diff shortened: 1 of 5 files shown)


mkopec (dismissed):

miczyg1:
But but but... now the keyboard backlight restoration does not work properly in Windows....

miczyg1:
I had to modify a few things to get the keyboard backlight be set up properly.  I had to modify the S76 EC device HID to make Linux not probe the driver, because the driver overrides the current backlight state set by ACPI. Moreover the driver uses 0-5 scale for white DAC backlight, while upstream EC switched to 0-ff range. So another incompatibility (of course this can be fixed on EC side, but I still dislike the idea of two backlight states firmware/ACPI plus Linux OS).

mkopec:
@miczyg1 with most recent changes, we only lose software control of keyboard backlight, right?

miczyg1:
@mkopec that's what I hope to happen, but I had no hardware at hand to verify it.

mkopec:
Tested on NV4x TGL in both S3 and s0ix modes, looks like things still work correctly

mkopec (approved):

PR #393, opened 2023-09-15
FUM iPXE autoboot

krystian-hebel:


miczyg1:
@krystian-hebel despite I have built a binary from this branch, iPXE was not automatically booted. Seems like BootNext did not work as intended.

miczyg1:
> I see, https://github.com/Dasharo/edk2/blob/dasharo/MdeModulePkg/Universal/BdsDxe/BdsEntry.c#L780 is probably the reason why. I'll see what can be done with it.

That might be the case.

krystian-hebel:
@miczyg1 https://github.com/Dasharo/edk2/pull/78 should fix that issue.

mkopec (approved):

PR #394, opened 2023-09-19
mb/clevo/tgl-u: add backlight levels in line with adl-p

mkopec:
Add the same linear backlight levels that we use for ADL-P models, in place of coreboot's default exponential levels.


mkopec:
@miczyg1 CI seems to be broken on this new branch

macpijan:
@mkopec We need to fix pre-commit here and then rebase: https://github.com/Dasharo/coreboot/pull/395#issuecomment-1727224537

PR #399, opened 2023-09-21
Clevo/ec power on ac

mkopec:
Companion EC PR: https://github.com/Dasharo/ec/pull/48

Uses the existing Power State After Failure option to control Power on AC feature. While not an ideal match, this lets us reuse existing code instead of adding even more EFI variables.


miczyg1:
I guess it is already included here: https://github.com/Dasharo/coreboot/pull/401

So closing.

PR #401, opened 2023-09-22
4.21 add missing features

mkopec:


src/vendorcode/dasharo/options.c
```diff
@@ -217,3 +217,36 @@ bool get_camera_option(void)
...
+
+	if (smmstore_lookup_region(&rdev))
+		return;
+
+	size = sizeof(*bat_cfg);
+	if (CONFIG(DRIVERS_EFI_VARIABLE_STORE))
+		ret = efi_fv_get_option(&rdev, &dasharo_system_features_guid, "BatteryConfig",
+					bat_cfg, &size);
+
+	if (ret != CB_SUCCESS || size != sizeof(*wdt_cfg)) {
```
mkopec, line 248:
```suggestion
	if (ret != CB_SUCCESS || size != sizeof(*bat_cfg)) {
```

src/mainboard/clevo/adl-p/ramstage.c
```diff
@@ -76,17 +76,56 @@ static void set_camera_enablement(void)
...
+		uint8_t index;
+		uint8_t value;
+	} __packed cmd = {
+		OPT_POWER_ON_AC,
+		0
+	};
+
+	cmd.value = dasharo_get_power_on_after_fail();
+
+	system76_ec_smfi_cmd(CMD_OPTION_SET, sizeof(cmd) / sizeof(uint8_t), (uint8_t *)&cmd);
```
miczyg1, line 101: Is it using the EC flash? If yes, then we should first check what is the current state from EC side and then if needed, issue CMD_OPTION_SET

miczyg1 (changes requested):

src/mainboard/clevo/adl-p/ramstage.c
mkopec: The check to avoid re-writing the same value to flash is done in the EC already: https://github.com/Dasharo/ec/pull/48/files#diff-603dee9b68e0fa20bb23508264232120edcd59f282a8dd59b999299c1a37f025R71

miczyg1:
@mkopec build fails

mkopec:
@miczyg1 build passes now

miczyg1 (approved):

miczyg1:
@mkopec can't fast-forward, could you please rebase?

mkopec:
@miczyg1 rebased

PR #403, opened 2023-09-23
[MSI] Add memory profile configuration

SergiiDmytruk:
Attempted to add simple recovery (reset memory profile option to JEDEC in romstage and restore it in ramstage), but writing a variable in romstage somehow ruined EFI variables storage.  I don't see why it would do that and left it out of the changes given that recover wasn't in requirements.


miczyg1 (commented):
LGTM, although there are change requests in the EDK2 and DasharoModulePkg

miczyg1 (approved):

PR #407, opened 2023-09-26
configs/config.protectli_vp2420: bump for Timeout fix

miczyg1:
UNTESTED.

I accidentally bricked the platform when developing fix and had to move to vp4670. So there may be some changes between EDK2 revision which break the booting on this platform.


miczyg1:
@mkopec tested and works now. Please review

miczyg1:
> @mkopec tested and works now. Please review

Ah the problem is not vboot itself but the amount of cache... This SoC have barely 4MB L3 cache, while FW_MAIN_A is nearly 3MB in size. Fetching it all and calculating a hash for vboot verifications fills up whole cache and romstage has no space for execution already... The issue was no triggered before because the payload was a bit smaller, but it doesn't;t change the fact that we were living on the edge here. I guess we will need a rebase and switch to VBOOT_CBFS_INTEGRATION to reduce the cache usage in pre-ram phase during verification.

miczyg1:
To reduce the cache usage I have removed separate verstage (now vboot will run inside botoblock)

mkopec (approved):

PR #408, opened 2023-09-27
Msi ms7d25/fpdt fix

miczyg1:


payloads/external/edk2/Kconfig
```diff
@@ -285,6 +285,17 @@ config EDK2_SETUP_PASSWORD
 	help
 	  Enables setup menu password in the TianoCore Payload.
 
+config EDK2_PERFORMANCE_MEASUREMENT_ENABLE
+	bool "Enable firmware performance measurement"
+	default n
+	help
+	  Enables firmware performance measurement in the TianoCore Payload.
+	  The measurements will be exposed to OS via FPDT ACPI table. Should be
+	  used in conjunction with EDK2_CPU_TIMER_LIB to match the coreboot
```
mkopec, line 294: maybe we can add a `depends on EDK2_CPU_TIMER_LIB` if it requires it to work?

miczyg1: Actually it doesn't. FPDT takes the timestamps in nanoseconds anyways, and this can be either returned via ACPI timer or TSC/CPU timer. However not sure if it will match with the ResetEnd timestamps that way... Also EDK2_CPU_TIMER_LIB is dangerous to use. If forcibly used on a platform that do not support certain CPUID leaf, then it will not boot. So I have left FPDT to be available and still have some measurements if wanted to.

mkopec: Ok, thanks

mkopec (approved):

PR #410, opened 2023-09-28
Protectli vault cml/cbfs serial UUID

miczyg1:


miczyg1:
> It looks like building without PerformanceLib fails

Yeah, guarded the libs unnecessarily. It is enough to guard the DXE module: https://github.com/Dasharo/edk2/commit/828979c450dabfe7d003dc7f3912dc99c97c04cc

mkopec (approved):

PR #412, opened 2023-09-29
New options (Camera, WiFi/BT, Console Redirection, Preinstalled OS boot options)

miczyg1:


miczyg1:
@mkopec we need this as well

PR #415, opened 2023-10-02
Msi ms7d25/sio power config

miczyg1:
Configures blinking LED on S3 sleep, PS/2 KBD and mouse wake from S3/S4. Also added a bit that is enabled by MSI firmware to handle power loss. Maybe this was causing issues reported during validation of the power status after power fail feature (at least it worked for me without any modifications).


mkopec (approved):
I'll look into enabling Wake on Serial, would be useful for mainboards with serial ports.

miczyg1:
> I'll look into enabling Wake on Serial, would be useful for mainboards with serial ports.

@mkopec note that it requires correct wiring the the RI# signals, which is not always present in the designs

PR #418, opened 2023-10-03
vendorcode/dasharo: add function to wipe options

mkopec:
Useful for boards that have the CMOS clear header simply connected to a GPIO


mkopec:
I wonder if this should cause a reset, or should this be handled somewhere else. If it calls a reset then it should also check if the region is already wiped in order to avoid bootloops.

miczyg1:
> I wonder if this should cause a reset, or should this be handled somewhere else. If it calls a reset then it should also check if the region is already wiped in order to avoid bootloops.

Depending on when the wipeout would be called, some options may have already been applied, so a reset is required IMO. 

Bootloops would not occur if we reset after the wipeout, because CMOS would already be sanitized

mkopec:
That's assuming that the CMOS clear request comes from a setting, in my use case it comes from a physical jumper / GPIO.

miczyg1:
> That's assuming that the CMOS clear request comes from a setting, in my use case it comes from a physical jumper / GPIO.

So it is a GPIO that tell BIOS to reset settings, instead of immediately clearing the CMOS? I am not sure I understand the purpose of such GPIO. Typically you have a CMOS clear jumper, which clears its memory when board is powered off. Then BIOS detects CMOS failure and clears the settings.

PR #419, opened 2023-10-05
NovaCustom release fixes

mkopec:


mkopec:
@miczyg1 no dice, even with https://github.com/Dasharo/coreboot/pull/419/commits/3d314fc6160451c7cd4a451e7d97bf59794bd963 nv4x tgl does not resume from S3. need to debug this properly.

PR #428, opened 2023-10-24
mb/clevo/adl-p: Add HDA verbs from vendor BIOS

miczyg1:


miczyg1:
@mkopec extracted from vendor BIOS.

I don't think it will help for headphones vs headset detection (which our customer reported for recent release), but may help with other things (like some weird sounds from speakers?).

UNTESTED

src/mainboard/clevo/adl-p/variants/nv40pz/hda_verb.c
```diff
@@ -6,7 +6,11 @@ const u32 cim_verb_data[] = {
 	/* Realtek, ALC256 */
 	0x10ec0256, /* Vendor ID */
 	0x15584041, /* Subsystem ID */
-	12, /* Number of entries */
+	18, /* Number of entries */
+	0x0205001a,
+	0x02048003,
+	0x0205001a,
+	0x0204c003,
```
mkopec, line 13: It would be nice if we could use a macro for these

mkopec (commented):
Yeah headset detection is a problem with the hardware, not sure if we can do anything from FW side.
Will test tomorrow.

src/mainboard/clevo/adl-p/variants/nv40pz/hda_verb.c
miczyg1: These are vendor specific verbs not standard azalia verbs, you would have to find datasheet to know what that means.

miczyg1:
> Yeah headset detection is a problem with the hardware, not sure if we can do anything from FW side. Will test tomorrow.

But maybe it will help with this: https://github.com/Dasharo/dasharo-issues/issues/224

mkopec:
AIUI that issue was caused by coil whine, I once tried enabling Acoustic Noise Optimization to mitigate this, but I didn't know what the correct setting should be there, so for now it's disabled. But it might help, I'll test.

miczyg1:
@mkopec any updates on testing it?

wessel-novacustom:
> AIUI that issue was caused by coil whine, I once tried enabling Acoustic Noise Optimization to mitigate this, but I didn't know what the correct setting should be there, so for now it's disabled. But it might help, I'll test.

It is not coil whine. Instead, it is really coming from the internal speakers. If we remove those internal speakers, the problem is no longer there. Also, this only happens on NV4x 11th Gen, not on NV4x 12th Gen.

Please let us know if these HDA verbs are included in the latest test binary we have received. If not, can it be?

miczyg1:
@mkopec I guess I will close this if these changes are unwanted...

mkopec (approved):

PR #429, opened 2023-10-26
Vp4 migrate

mkopec:
Migrated `vault_cml` directory from `protectli_vault_cml/develop`, cherrypicked dual FSP and TXT related commits and updated defconfigs. TXT config has ME enabled, regular config has ME disabled.
Not tested.


build.sh
```diff
@@ -66,14 +67,55 @@ function build_msi {
...
+		if [ -f protectli_blobs.zip ]; then
+			unzip protectli_blobs.zip -d 3rdparty/blobs/mainboard
+		else
+			echo "Platform blobs missing! You must obtain them first."
+			exit 1
+		fi
+	fi
+
+	version=$(git describe --abbrev=1 --tags --always --dirty)
+	version=${version//protectli_vault_cml_/}
```
miczyg1, line 93: Try to use this 

```
  DEFCONFIG="configs/config.${BOARD}"
  FW_VERSION=$(cat ${DEFCONFIG} | grep CONFIG_LOCALVERSION | cut -d '=' -f 2 | tr -d '"')
```

mkopec:
hmm, looks like with all these changes (dual FSP being the big one), UEFIPayload no longer fits in the 4MiB recovery region. need to do some size optimizations to make everything fit

mkopec:
I guess the bootsplash can be removed from UEFIPayload and placed in the persistent bootsplash FMAP region instead. No need to have it take up space in both recovery and RW_A slots.

build.sh
mkopec: Done: https://github.com/Dasharo/coreboot/pull/429/commits/8336bf206fd40e63a482e3a6e7cdc4dbe7a0796f

mkopec:
Removed bootsplash from edk2 and included it the BOOTSPLASH region instead. Payload size is reduced from 2032 to 1874KiB, leaving just 24KiB free.

macpijan:
> I guess the bootsplash can be removed from UEFIPayload and placed in the persistent bootsplash FMAP region instead. No need to have it take up space in both recovery and RW_A slots.

That one was left as a fallback in case some unsupported image gets loaded into the splash region IIUC?

macpijan:
> hmm, looks like with all these changes (dual FSP being the big one), UEFIPayload no longer fits in the 4MiB recovery region. need to do some size optimizations to make everything fit

We have not much more space to extend this I assume?
Size extension cause by double FSPs could be a valid reason to change the layout, if really necessary.

mkopec:
> That one was left as a fallback in case some unsupported image gets loaded (or is otherwise removed) into the splash region IIUC?

Right. Right now, the default TianoCore logo is shown when loading from the `BOOTSPLASH` region fails. This could be changed to something else, like a smaller Protectli logo, or a Dasharo logo.

> We have not much more space to extend this I assume?
> Size extension cause by double FSPs could be a valid reason to change the layout, if really necessary.

We could probably increase the recovery region a little, but right now RW section A is 4.6MiB so we only gain ~300KiB this way. The BOOTSPLASH region can also be shrinked, because it's hard to fill the entire 1MiB region with any compressed logo unless it's really complex or large.

mkopec:
@macpijan shrunk BOOTSPLASH by 512K and increased WP_RO by that amount. Removed unused CONSOLE region. Everything seems to fit fine now.

macpijan (approved):

PR #432, opened 2023-11-03
CI improvements

arturkow2:


.github/workflows/build.yml
```diff
@@ -62,12 +62,21 @@ jobs:
...
           make
+      - name: Create patched binary with pre-enabled serial output
+        run: |
+          cp build/coreboot.rom build/coreboot_serial_enabled.rom
+          build/util/cbfstool/cbfstool build/coreboot_serial_enabled.rom write -r SMMSTORE -f ci/smmstore-serial-enabled
+
+          if [ -e "ci/${{ matrix.vendor }}_${{ matrix.model }}_pretrained" ]; then
+            cp build/coreboot_serial_enabled.rom build/coreboot_serial_enabled_pretrained.rom
+            build/util/cbfstool/cbfstool build/coreboot_serial_enabled_pretrained.rom write -r RW_MRC_CACHE -f ci/${{ matrix.vendor }}_${{ matrix.model }}_pretrained
+          fi
```
miczyg1, line 73: Why do we need a pretrained binary? This is not a stable solution.

.github/workflows/build.yml
```diff
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         vendor: [ msi ]
-        model: [ ms7d25_ddr4, ms7d25_ddr5, ms7e06_ddr4, ms7e06_ddr5 ]
```
miczyg1, line 48: I know it is only for testing, but we need all platforms to be built.

.github/workflows/build.yml
```diff
@@ -100,3 +109,60 @@ jobs:
       - name: Build Dasharo
         run: |
           ./build.sh ${{ matrix.model }}
+  test_msi:
+    runs-on: self-hosted
+    strategy:
+      matrix:
+        vendor: [ msi ]
+        model: [ ms7d25_ddr5 ]
```
miczyg1, line 117: We have two more variants to test

miczyg1 (changes requested):

.github/workflows/build.yml
arturkow2: Got this accidentially into PR, was only to speed up tests.

.github/workflows/build.yml
arturkow2: @miczyg1 We need to enable serial port but can do this only by manually entering setup and enabling it. This was workaround suggested by @macpijan : https://jira-3mdeb.atlassian.net/browse/DSH-762?focusedCommentId=133640

macpijan: I suggested just enabling serial, not including pretraineded region.

.github/workflows/build.yml
macpijan: Does it mean it has to be all in a single MR? Why not merge a single board first to see how it works in practice, and add more later on? We will be able to gather problems/failures and could discuss how to resolve these already.

We've asked to move from github actions anyway, so maybe we will do this as well then.

macpijan: Then I'd ask you to improve it, please, to meet your requirements.

.github/workflows/build.yml
miczyg1: Removed: https://github.com/Dasharo/coreboot/pull/432/commits/bf4b3d93192b5778c9a0d6ca75b7a5edf4cb8228#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L70-L73

.github/workflows/build.yml
miczyg1: Added all models we have in lab: https://github.com/Dasharo/coreboot/pull/432/commits/bf4b3d93192b5778c9a0d6ca75b7a5edf4cb8228#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R112

ci/ci.sh
```diff
@@ -0,0 +1,150 @@
...
+        exit 1
+        ;;
+esac
+
+options+=(
+    -v device_ip:$device_ip
+    -v rte_ip:$rte_ip
+    -v pikvm_ip:$pikvm_ip
+    -v config:$config
+    -v snipeit:no
```
miczyg1, line 82: This is a big problem that CI does not use snipeit. Any push to a PR will cause an attempt to access hardware resources that not necessarily are free. I can already imagine a pile of problems arising from this. Why CI can't use sniepit?

miczyg1 (changes requested):

ci/ci.sh
macpijan: It should, let's try to just remove it. It used to fail when the asset was not available like this: https://github.com/Dasharo/open-source-firmware-validation/actions/runs/6629204171/job/18007955133#step:5:158

Also, not ideal, but at least it does not break things, which is a nice starting point.

Right now it cannot reach snipeit it seems: https://github.com/Dasharo/open-source-firmware-validation/actions/runs/7385294785/job/20089770703#step:5:117 which may be infrastructure problem.

miczyg1: Failed jobs can always be re-run when needed. Besides it would be much more safe solution.

macpijan: > Was our runner even prepared to access sniepit? I believe it needs https://github.com/Dasharo/osfv-scripts/tree/main/osfv_cli to be installed and proper entry in /etc/hosts to be present...

Yes, it was. I have linked you a job where it reached it correctly.

mkopec: `/etc/hosts` entries should no longer be needed, IT added it to DNS

miczyg1: I have re-run the failed jobs. However the Z690 DR5 did not fail for some reason and is not restarted...

miczyg1: I had bad names of the platform configs in the ci.sh too it seems. Let's see how it will look now

PR #433, opened 2023-11-13
Nvc tbt fixes

mkopec:
Hide the TBT DMA devices so Windows stops complaining about missing drivers. Verify that XDomain networking still works under Linux. ATM it's not possible to verify if other USB4 tunneling functionality is working.

Needs https://github.com/Dasharo/ec/pull/52

### DIFF
--- a/src/ec/system76/ec/acpi/ec.asl
+++ b/src/ec/system76/ec/acpi/ec.asl
@@ -96,6 +96,20 @@ Device (\_SB.PCI0.LPCB.EC0)
 		}
 	}
 
+	Method (S0IX, 1, Serialized) {
+		Printf ("EC: S0ix hook")
+		If (ECOK) {
+			S0XH = Arg0
+		}
+	}
+
+	Method (EDSX, 1, Serialized) {
+		Printf ("EC: Display hook")
+		If (ECOK) {
+			DSPH = Arg0
+		}
+	}
+
 	Method (_Q0A, 0, NotSerialized) // Touchpad Toggle
 	{
 		Printf ("EC: Touchpad Toggle")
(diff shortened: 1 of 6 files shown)


wessel-novacustom:
I can confirm that the warning sign in device manager at the Thunderbolt Controller is gone now (NS70PU-i7). On that device, sleep mode is working now, but it takes more than 30 seconds.

I can also confirm that there is no such a warning sign under Windows 11 for NV4x-TGL, but this hasn't been confirmed before to be a problem/bug on that device.

macpijan:
I have picked edc2df262494a56e4a7000cd48007787062bcfd7 into our integration branch, so we can take advantage of building test binaries in the CI

wessel-novacustom:
Tomorrow, we will need to try if Thunderbolt is working properly on all devices.

src/soc/intel/alderlake/acpi/tcss.asl
```diff
@@ -690,7 +690,7 @@ Scope (\_SB.PCI0)
 		Method (_STA, 0x0, NotSerialized)
 		{
 			If (DME0 == 1) {
-				Return (0x0F)
+				Return (0x0B)
```
miczyg1, line 693: This is not a good solution. We just hide errors, not missing drivers. When booting Windows without a dock, al TBT related devices are working properly without yellow bangs.

mkopec: I had issues where it either could not find drivers for the devices, or it found them but they errored out with code 43, but I understand. Will revert.

mkopec: Reverted in 1dccfab207a274aab62cfe6f32666f7f30b39f46

PR #434, opened 2023-11-14
src/mb/clevo: Avoid conflict of VBNV offset and ramtop in CMOS

miczyg1:
Should fix: https://github.com/Dasharo/dasharo-issues/issues/556

However, I could not reproduce the popup after the upgrade. But most likely this is the culprit.


mkopec:
@miczyg1 the checksum ranges also need to be updated, right?

macpijan:
I rebased to enable CI

miczyg1:
> @miczyg1 the checksum ranges also need to be updated, right?

I don't see a need

mkopec:
@miczyg1 

> I don't see a need

The vbnv must be outside of the checksummed range, otherwise coreboot may reset it, which can render to bootloop in the recovery path

https://github.com/Dasharo/coreboot/pull/251
https://github.com/Dasharo/dasharo-issues/issues/242

PR #435, opened 2023-11-14
Windows11 shutdown fix

miczyg1:


miczyg1:
Fixes: https://github.com/Dasharo/dasharo-issues/issues/562

macpijan:
I rebased to enable CI build

mkopec:
@miczyg1 could you summarize here what are the effects of switching to SCM on Linux?

miczyg1:
> @miczyg1 could you summarize here what are the effects of switching to SCM on Linux?

Dock is still working on ADL as intended

PR #450, opened 2023-12-21
payloads/iPXE: add TRUST_CMD switch

macpijan:
Upstream tracking: https://review.coreboot.org/c/coreboot/+/79684

Change-Id: I4dca93451d16122a1c902bacc7354298d3a1ae52


macpijan:
Tested up to the point new commands appearing in the .efi file via `strings` (previously not there.


I think for Dasharo, we want this set by default, so we do not lose it in some configs.

mkopec (approved):

PR #456, opened 2024-01-22
[WIP] EC FW sync

mkopec:
Change-Id: I7f2c9f756102ad6e63d8aafd3013003d31676ce4

### DIFF
--- a/src/ec/system76/ec/system76_ec.c
+++ b/src/ec/system76/ec/system76_ec.c
@@ -244,10 +244,11 @@ static uint32_t flash_read(uint8_t *dest, uint32_t len) {
 
 		/* Check how many data bytes were actually read */
 		bytes_read = system76_ec_read(REG_DATA + 1);
+		return bytes_read;
 
 		/* Sanity check */
-		if (bytes_read == 0 || bytes_read > SYSTEM76_EC_SIZE - 4)
-			return 0;
+		if (bytes_read > SYSTEM76_EC_SIZE - 4)
+			break;
 
 		/* Read data bytes, index should be valid due to length test above */
 		for (i = 0; i < bytes_read; i++)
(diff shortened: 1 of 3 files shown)


mkopec:
So far it can compare versions:

```
[DEBUG]  EC update found (46113 bytes)
[DEBUG]  EC update target:  clevo/nv40mz
[DEBUG]  EC update version:  2023-12-07_bf0073c
[DEBUG]  EC current version: 2023-12-08_2b2c17a
[DEBUG]  EC update required!
```

src/ec/system76/ec/system76_ec.c
```diff
@@ -148,3 +187,156 @@ void system76_ec_set_bat_threshold(enum bat_threshold_type type, uint8_t value)
...
+		0,
+	};
+
+	/* If we failed to jump to scratch ROM, then we can probably continue booting. */
+	if (system76_ec_smfi_cmd(CMD_SPI, sizeof(smfi_cmd) / sizeof(uint8_t), (uint8_t *)&smfi_cmd)) {
+		printk(BIOS_ERR, "EC update failed!\n");
+		return;
+	}
+
+	uint8_t current_rom[SPI_FLASH_SIZE] = { 0xFF };
```
miczyg1, line 328: Don't do it like that. You basically request 128K of memory to be allocated on stack. coreboot's stake is only a couple of kilobytes...

Do a malloc instead and free the buffer after use. Then the buffer will be used as heap.

@mkopec

mkopec: makes sense, applied your suggestion in https://github.com/Dasharo/coreboot/pull/456/commits/e34fee6c8657ee449fdf5824651ad457f7c81f01 but it fails as well with this error:

```
memalign(boundary=8, size=131072): failed: Tried to round up free_mem_ptr 0x76b0ba30 to 0x76b2ba30
but free_mem_end_ptr is 0x76b1ba30
```

I suppose we don't actually need to allocate the full 128KB, for verification after update we can just compare chunk by chunk

miczyg1: As you can see, coreboot heap is 64KiB too short to allocate full flash size.

> I suppose we don't actually need to allocate the full 128KB, for verification after update we can just compare chunk by chunk

Probably better idea.

PR #458, opened 2024-01-30
Ifdtool HAP bit

miczyg1:


util/ifdtool/ifdtool.c
```diff
@@ -1506,17 +1506,219 @@ static void set_pchstrap(struct fpsba *fpsba, const struct fdbar *fdb, const int
...
+	case PLATFORM_SKLKBL:
+		/* No exceptions, always strap 0 */
+		return 0;
+	case PLATFORM_IFD2:
+		printf("To change the AltMeDisable/HAP bit please provide the platform as parameter\n");
+		return -1;
+	case PLATFORM_CNL:
+		/*
+		 * This gets little tricky here.
+		 * 1. CNL/CFL/CML-LP strap 106 (FPSBA + 0x70), but
```
mkopec, line 1521: Would it not make more sense to pass the platform name / variant / SKU as an argument to ifdtool at this point?

util/ifdtool/ifdtool.c
```diff
@@ -1506,17 +1506,219 @@ static void set_pchstrap(struct fpsba *fpsba, const struct fdbar *fdb, const int
...
+			if (((fpsba->pchstrp[0x70 / 4] >> 4) & 0xfff) == 0x1b) {
+				printf("Detected Cannon Lake/Coffee Lake/Comet Lake LP "
+				       "platform flash descriptor\n");
+				return 0x70 / 4;
+			}
+		}
+		printf("Unusual Cannon Lake/Coffee Lake/Comet Lake flash descriptor detected\n"
+		       "Not modifying flash descriptor\n");
+		return -1;
+	case PLATFORM_TGL:
```
mkopec, line 1564: Or revise the platform names, since right now one can refer to multiple different platform codenames like TGL / RKL here

util/ifdtool/ifdtool.c
miczyg1: I didn't want to redesign ifdtool that much. Besides it is for dumping and modifying FD in scope of region boundaries, flash master permissions etc. Having support for HAP is an extra nice-to-have.

util/ifdtool/ifdtool.c
miczyg1: coreboot does not even have support for RKL so adding a dedicated platform enum for it doesn't seem beneficial. Moreover basically RKL is the same as TGL-H. AS long as the code does what it needs to do, it is fine for me (and probably for many others as well).

mkopec (dismissed):

miczyg1:
@mkopec I also have to add support for dumping the state in `dump_fpsba`

miczyg1:
@mkopec please have a look at https://github.com/Dasharo/coreboot/pull/458/commits/a667f4632b1821a7270f261c572f48ba5626b16d

Now the HAP state should be printed properly when `-d` switch is used

PR #461, opened 2024-02-14
Msi heads

miczyg1:


miczyg1:
The situation on MSI is a little bit tricky. Now that the BIOS region is 28MiB there are few problems with that:

1. coreboot tries to place FIT at the lowest address possible in flash, which will end up outside of the allowed 16 MiB range, so I forcefully allocate it manually.
2. coreboot tries to place microcode at the lowest address possible in flash, which will end up outside of the allowed 16 MiB range, so I forcefully allocate it manually.
3. There should be no region that crosses the boundary of fixed memory mapped flash space and extended memory mapped flash space. Makefiles should catch it, but for some reason they don't. Anyway, to avoid any CBFS file cross that boundary, I am placing one of the files exactly at 16MiB boundary (0xff000000 + CBFS file metadata size)

We wouldn't have any of those problems if:

1. We didn't use FlashBIOS (ROMHOLE is right in the middle of fixed 16MiB memory mapping windows, so can;t fit heads neither before nor after the ROMHOLE, kernel is just too big)
2. FlashBIOS could accept ROMHOLE in any other place than address 0x17c0000 in flash. Although the magic strings indicate the address of ROMHOLE, it doesn't seem like the uC responsible for flashing can cope with different address

miczyg1:
@mkopec your review is needed here

src/mainboard/msi/ms7d25/Makefile.inc
```diff
@@ -11,3 +11,22 @@ ramstage-y += smbios.c
...
+bpa.bin-type := raw
+bpa.bin-compression := none
+bpa.bin-position := 0xff080000 # Fixed position from original MSI firmware
+
+# Workaround: place build_info exactly at the EXT_BIOS_WINDOW boundary (plus
+# the metadata size) to avoid files getting fragmented between
+# FIXED_BIOS_WINDOW and EXT_BIOS_WINDOW.
+build_info-COREBOOT-position := 0xff000040
+
+endif
```
mkopec, line 32: missing EOL?

util/msi/romholetool/romholetool.c
```diff
@@ -158,21 +158,26 @@ struct msi_romhole {
...
 	printf("romholetool: Create ROMHOLE for MSI BIOS updates compatibility\n");
-	printf("Usage: romholetool [options] -i build.h  -l fmap_config.h -o <filename>\n");
-	printf("-l | --layout <FILE>             The fmap_config.h file to parse\n");
-	printf("-o | --output <FILE>             The file to generate\n");
-	printf("-h | --help                      Print this help\n");
+	printf("Usage: romholetool [options] -i build.h -l fmap_config.h -o <filename>\n");
+	printf("       romholetool [options] -i build.h -a 0xff7c0000 -s 0x20000 -o <filename>\n");
+	printf("-l | --layout  <FILE>             The fmap_config.h file to parse\n");
+	printf("-a | --address <HEX>              The address of ROMHOLE in CBFS\n");
+	printf("-s | --size    <HEX>              The address of ROMHOLE in CBFS\n");
```
mkopec, line 178:
```suggestion
	printf("-s | --size    <HEX>              The size of ROMHOLE in CBFS\n");
```

mkopec (changes requested):

util/msi/romholetool/romholetool.c
miczyg1: https://github.com/Dasharo/coreboot/pull/461/commits/8affdcf545ca134856b1e24b984a67672ca501ca#diff-6fcf9af3dec0719700c1bb178bc3f3ddae869efdcbd67b7f5a6755003374a03aR178

src/mainboard/msi/ms7d25/Makefile.inc
miczyg1: https://github.com/Dasharo/coreboot/pull/461/commits/d71eaf3228791e4c2308b6045bdfef73ab8315ca#diff-0c5063207cef14f59aa8298ca57a6c579e9b3609481dc8a389f9922fed88baf5R32

mkopec (approved):

PR #463, opened 2024-02-15
Protectli new features

miczyg1:


src/security/lockdown/Kconfig
```diff
@@ -62,7 +62,6 @@ config BOOTMEDIA_LOCK_WHOLE_NO_ACCESS
 
 config BOOTMEDIA_LOCK_WPRO_VBOOT_RO
 	bool "Write-protect WP_RO FMAP region in boot medium"
-	depends on VBOOT
```
mkopec, line 65: Why is this needed? Won't `BOOTMEDIA_LOCK_WHOLE_RO` with the old fmap do what we want here? I'm assuming the point is to write-protect COREBOOT and FMAP without enabling Vboot entirely

miczyg1:
@mkopec suggestion applied. Also rebased on the most recent changes on dasharo branch and set the EDK2 rev to the final hash after merging the PRs.

mkopec (dismissed):

miczyg1:
@mkopec I think I am missing some DEC file somewhere thus it is not building. I will push the fix directly, seems like a very minor issue

mkopec:
Sounds good

miczyg1:
The problem was the scoping... I cannot use DasharoModulePkg PCDs straight in UefiPayloadPkg.dsc, but only in modules that include DasharoModulePkg.dec...

PR #465, opened 2024-02-29
clevo: enable secdata mocking

mkopec:
There are issues with secdata on these platforms and it's been reported to cause erroneous recovery entry: https://github.com/Dasharo/dasharo-issues/issues/556

Turn on secdata mocking to disable TPM usage for secdata.


miczyg1 (approved):
That basically means vboot antirollback will be disabled

PR #472, opened 2024-03-14
Acpi wdat

mkopec:


mkopec:
Basic functionality is working: reading min / max timeout via sysfs, pinging the watchdog to prevent a reset. An OS with wdat_wdt driver loaded does not reboot after the default timeout expires, and an os without the driver does reset. Reading / writing the timeout does not work correctly yet and I'm not sure if the state of the watchdog (running/stopped) is correct.

mkopec:
Ah, needed to add preserve flag to the ping command.

mkopec:
Added storing bootstatus in the scratch bits of the OC_WDT_CTL register. This lets WDAT read the bootstatus flag regardless of the ICC survavibility mode.

mkopec:
Enabled for VP46xx only for now. When WDAT aware OS like Archlinux or Windows is booted, no resets happen as long as the OS is not frozen.

src/soc/intel/common/tco.h
```diff
@@ -0,0 +1,47 @@
...
+#define TCO2_STS			0x06
+#define  TCO2_STS_SMLINK_SLAVE_SMI	(1 << 2)
+#define  TCO2_STS_SECOND_TO		(1 << 1)
+#define  TCO2_INTRD_DET			(1 << 0)
+#define TCO1_CNT			0x08
+#define  TCO1_LOCK			(1 << 12)
+#define  TCO1_TMR_HLT			(1 << 11)
+#define  TCO1_NMI2SMI_EN		(1 << 9)
+#define  TCO1_NMI_NOW			(1 << 8)
+#define TCO2_CNT			0x0A
```
macpijan, line 30: Some mixup in indentation?

mkopec: Single indentation for registers, double indentation for bitfields :)

mkopec: My changes are basically just these commits: https://github.com/Dasharo/coreboot/pull/472/files/78faf9f9c23ceaa09e24067b3b652535888e74b4..da615616e7459def116a2daecb63e8a16f0b4b0f

macpijan: Got confused by seeing this, thanks for clarification :upside_down_face:

mkopec:
Also enabled on ADL-P and EHL platforms.

src/mainboard/protectli/vault_adl_p/Kconfig
```diff
@@ -3,17 +3,19 @@ if BOARD_PROTECTLI_VP66XX
...
 	select HAVE_ACPI_RESUME
 	select HAVE_ACPI_TABLES
 	select INTEL_GMA_HAVE_VBT
 	select MEMORY_MAPPED_TPM
+	select SOC_INTEL_ALDERLAKE_PCH_P
+	select SOC_INTEL_COMMON_BLOCK_HDA_VERB
+	select SOC_INTEL_COMMON_OC_WDT_ENABLE
+	select SOC_INTEL_COMMON_OC_WDT_WDAT
+	select SOC_INTEL_ENABLE_USB4_PCIE_RESOURCES
+	select SUPERIO_ITE_IT8659E
```
miczyg1, line 18: Why do we have to change the order of options here? SOC_INTEL_ALDERLAKE_PCH_P is first for a reason, it selects most of the microarchitecture specific options.

`SOC_INTEL_COMMON_OC_WDT_ENABLE` this should not be selected. It means we cannot disable the watchdog.
`SOC_INTEL_COMMON_OC_WDT` is selected by ADL_P already.

src/mainboard/protectli/vault_cml/Kconfig
```diff
@@ -3,19 +3,21 @@ if BOARD_PROTECTLI_VP46XX
 config BOARD_SPECIFIC_OPTIONS
```
miczyg1, line 3: a/a don't reorder please

src/mainboard/protectli/vault_ehl/Kconfig
```diff
@@ -4,15 +4,17 @@ if BOARD_PROTECTLI_VP2420
...
-	select SPI_FLASH_MACRONIX
 	select INTEL_GMA_HAVE_VBT
-	select SEABIOS_ADD_SERCON_PORT_FILE if PAYLOAD_SEABIOS
 	select MEMORY_MAPPED_TPM
+	select SEABIOS_ADD_SERCON_PORT_FILE if PAYLOAD_SEABIOS
+	select SOC_INTEL_COMMON_OC_WDT_ENABLE
+	select SOC_INTEL_COMMON_OC_WDT_WDAT
+	select SOC_INTEL_ELKHARTLAKE
+	select SPI_FLASH_MACRONIX
+	select SUPERIO_ITE_IT8613E
```
miczyg1, line 17: a/a don't reorder please

src/soc/intel/common/block/oc_wdt/acpi.c
```diff
@@ -0,0 +1,156 @@
...
+#define OC_WDT_MIN_VALUE	0
+#define OC_WDT_MAX_VALUE	1023
+#define OC_WDT_PERIOD_MS	1000
+
+static bool fill_wdat_timeout_entry(acpi_wdat_entry_t *entry)
+{
+	memset((void *)entry, 0, sizeof(acpi_wdat_entry_t));
+
+	entry->action = ACPI_WDAT_SET_COUNTDOWN;
+	entry->instruction = ACPI_WDAT_WRITE_COUNTDOWN | ACPI_WDAT_PRESERVE_REGISTER;
```
miczyg1, line 16: Bits 24 and 25 are write-1-clear, so they shouldn't be preserved. If OS writes 1 to these, they will be cleared.

src/soc/intel/common/block/oc_wdt/acpi.c
```diff
@@ -0,0 +1,156 @@
...
+	entry->mask = PCH_OC_WDT_CTL_TOV_MASK;
+	entry->register_region.space_id = ACPI_ADDRESS_SPACE_IO;
+	entry->register_region.addrl = PCH_OC_WDT_CTL;
+	entry->register_region.access_size = ACPI_WDAT_ACCESS_SIZE_DWORD;
```
miczyg1, line 20: Ste access size to word (16 bits)  as we don't need to write whole register. That way we can also avoid clearing status bits

src/soc/intel/common/block/oc_wdt/acpi.c
```diff
@@ -0,0 +1,156 @@
...
+{
+	memset((void *)entry, 0, sizeof(acpi_wdat_entry_t));
+
+	entry->action = action;
+	entry->instruction = instruction;
+	entry->value = value;
+	entry->mask = PCH_OC_WDT_CTL_EN;
+	entry->register_region.space_id = ACPI_ADDRESS_SPACE_IO;
+	entry->register_region.addrl = PCH_OC_WDT_CTL;
+	entry->register_region.access_size = ACPI_WDAT_ACCESS_SIZE_DWORD;
```
miczyg1, line 52: Same here, set it to word access, PCH_OC_WDT_CTL_EN is at bit 14.

src/soc/intel/common/block/oc_wdt/acpi.c
```diff
@@ -0,0 +1,156 @@
...
+{
+	memset((void *)entry, 0, sizeof(acpi_wdat_entry_t));
+
+	entry->action = ACPI_WDAT_RESET;
+	entry->instruction = ACPI_WDAT_WRITE_VALUE | ACPI_WDAT_PRESERVE_REGISTER;
+	entry->value = PCH_OC_WDT_CTL_RLD;
+	entry->mask = PCH_OC_WDT_CTL_RLD;
+	entry->register_region.space_id = ACPI_ADDRESS_SPACE_IO;
+	entry->register_region.addrl = PCH_OC_WDT_CTL;
+	entry->register_region.access_size = ACPI_WDAT_ACCESS_SIZE_DWORD;
```
miczyg1, line 67: To avoid clearing the status bits, we can only write the most significant bit in the PCH_OC_WDT_CTL without preserving it, i.e. make the offset of PCH_OC_WDT_CTL + 3 be written with 0x80 (PCH_OC_WDT_CTL_RLD properly shifted)

src/soc/intel/common/block/oc_wdt/acpi.c
```diff
@@ -0,0 +1,156 @@
...
+	if (!fill_wdat_boot_status_entry(entry, ACPI_WDAT_GET_STATUS,
+					 ACPI_WDAT_READ_VALUE, 1))
+		goto out_err;
+
+	entry++;
+
+	/* Set boot status */
+	if (!fill_wdat_boot_status_entry(entry, ACPI_WDAT_SET_STATUS,
+					 ACPI_WDAT_WRITE_VALUE | ACPI_WDAT_PRESERVE_REGISTER,
+					 0))
```
miczyg1, line 106: This should also be handled in a way that the Status bits are not cleared if they don't need to be

miczyg1 (changes requested):

src/mainboard/protectli/vault_adl_p/Kconfig
mkopec: Just alphabetized them. Will revert.

mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/6206afd8b85a7656a109566cc1e5f6d2c39e9674

src/mainboard/protectli/vault_cml/Kconfig
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/6206afd8b85a7656a109566cc1e5f6d2c39e9674

src/mainboard/protectli/vault_ehl/Kconfig
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/6206afd8b85a7656a109566cc1e5f6d2c39e9674

src/mainboard/protectli/vault_adl_p/Kconfig
mkopec: >  SOC_INTEL_COMMON_OC_WDT is selected by ADL_P already.

Here I'm selecting ` SOC_INTEL_COMMON_OC_WDT_WDAT` (maybe the naming is unfortunate)

src/soc/intel/common/block/oc_wdt/acpi.c
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/f99cf95497da3151fe21dc6806c23af9b32e5129

src/soc/intel/common/block/oc_wdt/acpi.c
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/f99cf95497da3151fe21dc6806c23af9b32e5129

src/soc/intel/common/block/oc_wdt/acpi.c
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/f99cf95497da3151fe21dc6806c23af9b32e5129

src/soc/intel/common/block/oc_wdt/acpi.c
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/f99cf95497da3151fe21dc6806c23af9b32e5129

src/soc/intel/common/block/oc_wdt/acpi.c
mkopec: https://github.com/Dasharo/coreboot/pull/472/commits/f99cf95497da3151fe21dc6806c23af9b32e5129

mkopec:
Need to re-test after the access width changes.

src/soc/intel/common/block/oc_wdt/acpi.c
mkopec: I knew that happens, but I wasn't sure if it was OK to access the register at lower access widths. Will test with your suggestions and report back

src/mainboard/protectli/vault_adl_p/Kconfig
mkopec: > SOC_INTEL_COMMON_OC_WDT_ENABLE this should not be selected. It means we cannot disable the watchdog.

@miczyg1 are you sure about that?

https://github.com/Dasharo/coreboot/blob/6b4a0b81635b919c6fa21814b3b95c7612fe5a6f/src/soc/intel/common/block/oc_wdt/oc_wdt.c#L95
https://github.com/Dasharo/coreboot/blob/6b4a0b81635b919c6fa21814b3b95c7612fe5a6f/payloads/external/edk2/Makefile#L248-L251

mkopec:
@miczyg1 Appears to still work correctly after the latest changes. However, the watchdog is now enabled by default, it probably shouldn't be, as some users may be using an OS without watchdog support OOTB (Ubuntu doesn't load the module automatically for some reason, for example)

src/mainboard/protectli/vault_adl_p/Kconfig
miczyg1: What I mean is `select SOC_INTEL_COMMON_OC_WDT_ENABLE` will prevent you from chanign this option in the menuconfig (it will be forcibly selected as enabled and attempts to disable it will not work)

src/soc/intel/common/block/oc_wdt/acpi.c
miczyg1: As long as your accesses are aligned i.e. word access on word boundary, it should be fine. Byte access has no resctrictions.

src/mainboard/protectli/vault_adl_p/Kconfig
miczyg1: An acceptable solution would be

```
config SOC_INTEL_COMMON_OC_WDT_ENABLE
       default y
```

But thas shadows the options being set in configs/config.protectli* files, so you cannot early determine what is enabled and what not, unless you look deeper into the code, which I am not in favor of...

mkopec: Oh I see what you mean. Will fix

mkopec: Fixed in https://github.com/Dasharo/coreboot/pull/472/commits/6df14ccf6f0782300edb8a057b2067088193d15f

mkopec:
Had to make some changes on the edk2 side too: https://github.com/Dasharo/edk2/pull/124 https://github.com/Dasharo/DasharoModulePkg/pull/42

src/acpi/acpi.c
```diff
@@ -1166,6 +1166,30 @@ unsigned long acpi_gtdt_add_watchdog(unsigned long current, uint64_t refresh_fra
...
 }
 
+__weak unsigned long acpi_soc_fill_wdat(acpi_wdat_t *wdat, unsigned long current)
+{
+	return 0;
+}
+
+static void acpi_create_wdat(acpi_header_t *header, void *unused)
+{
+	if (!CONFIG(ACPI_WDAT_WDT))
```
miczyg1, line 1176: Should we create the table when watchdog is disabled?

mkopec: If neither `CONFIG_SOC_INTEL_COMMON_OC_WDT_WDAT` nor `SOC_INTEL_COMMON_BLOCK_TCO_WDT_WDAT` are selected then we should probably not generate the table, yes. I guess `ACPI_WDAT_WDT` should depend on a backend WDT being selected for it?

miczyg1: @mkopec I meant UEFI variable

mkopec: Hmm, in that case I think we should, the firmware won't start the watchdog, but the OS can still decide to use it

miczyg1: >  the OS can still decide to use it

What about user decision?

mkopec: I meant that the user can configure the OS to use or not to use the watchdog :) I don't see how hiding the WDAT table from the OS is useful

miczyg1: Okay, if it is possible then all good.

miczyg1:
Needs a rebase. Do we have some test logs maybe?

mkopec:
@miczyg1 I posted some logs here: https://github.com/Dasharo/open-source-firmware-validation/pull/229

Probably will need to re-run after rebasing

miczyg1 (approved):

mkopec:
@miczyg1 Had to switch back to 32bit register accesses, turns out word and byte accesses don't work correctly on alderlake at least https://github.com/Dasharo/coreboot/pull/472/commits/86cc5bcaea9735b92dc9c8ff5ea4cd69ec226496

miczyg1 (dismissed):

mkopec:
@miczyg1 rebased on dasharo-24.02.1

miczyg1 (approved):

PR #473, opened 2024-03-26
Bay Trail TXE Secure Boot

miczyg1:


miczyg1:
FSP approach is unfeasible. Closing.

PR #476, opened 2024-04-03
Vp66xx full build

miczyg1:
One caveat is that with new ME, it does not accept soft disable HECI command and casueda resetl oop. I had to use HAP to disable it.


macpijan (approved):

macpijan:
Rebased and ff-merged. Thanks.

PR #482, opened 2024-04-12
meteorlake inteltool backup

filipleple:


macpijan:
I suggest we push to gerrit right away @filipleple . I assume we might not be the only ones interested in this.

PR #484, opened 2024-04-15
mb/clevo/tgl-u: update VBT for FSP A.0.7E.70

mkopec:
Generated from TigerLakeFspBinPkg/Client/SampleCode/Vbt/Vbt.bin using DisCon 3.6

Change-Id: I14ca3ae6fd63d95637704d9be16660336b9248da


mkopec:
@miczyg1 yes, this fixed display init on tgl. Some offsets changed between (previous VBT format version was 236, current is 250)

PR #505, opened 2024-05-14
Add CPU throttling threshold option support for TGL and ADL

filipleple:


src/mainboard/clevo/adl-p/devtree.c
```diff
@@ -0,0 +1,24 @@
...
+void devtree_update(void)
+{
+	config_t *cfg = config_of_soc();
+
+	// read the CpuThrottlingThreshold efivar
+	uint8_t CpuThrottlingThreshold;
+	CpuThrottlingThreshold = get_cpu_throttling_threshold();
+
+
+	// read TjMax
```
mkopec, line 20: This can probably be defined statically as a Kconfig option for each CPU generation, I think.

mkopec:
I've merged https://github.com/Dasharo/coreboot/pull/513 which is a superset of this PR, so this it can be closed I think

PR #512, opened 2024-05-28
mainboard/protectli/vault_ehl: Disable HWP

miczyg1:
HWP is too aggressive in power savings and does not let using full bandwidth of Ethernet controllers without additional stressing of the CPUs (2Gb/s vs 2.35Gb/s with stressing, measured with iperf3). Let the Linux use acpi-cpufreq governor driver instead of intel_pstate by disabling HWP.

### DIFF
--- a/src/mainboard/protectli/vault_ehl/mainboard.c
+++ b/src/mainboard/protectli/vault_ehl/mainboard.c
@@ -12,6 +12,14 @@ static void mainboard_final(void *chip_info)
 
 void mainboard_silicon_init_params(FSP_S_CONFIG *silconfig)
 {
+	/*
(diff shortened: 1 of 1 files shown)


mkopec:
@miczyg1 did you try tweaking the energy performance bias value? default is 6 which could be restricting performance

miczyg1:
> @miczyg1 did you try tweaking the energy performance bias value? default is 6 which could be restricting performance

I tried everything, EPB, EPP and it simply did not reach the expected bandwidth without additional stressing.

mkopec (approved):

macpijan (approved):

PR #518, opened 2024-06-12
Downcoring and HT options

miczyg1:


src/soc/intel/alderlake/romstage/fsp_params.c
```diff
@@ -180,6 +180,18 @@ static void fill_fspm_cpu_params(FSP_M_CONFIG *m_cfg,
 	m_cfg->PrmrrSize = get_valid_prmrr_size();
 	m_cfg->EnableC6Dram = config->enable_c6dram;
 	m_cfg->HyperThreading = get_uint_option("hyper_threading", CONFIG(FSP_HYPERTHREADING));
+
+	if (CONFIG(DRIVERS_EFI_VARIABLE_STORE) && CONFIG(FSP_HYPERTHREADING))
```
SergiiDmytruk, line 184: Maybe `get_uint_option("hyper_threading", CONFIG(FSP_HYPERTHREADING))` should simply be used as a default value in `get_hyper_threading_option()`?  Should make it easier to reuse for other SOCs.

src/soc/intel/alderlake/romstage/fsp_params.c
```diff
@@ -180,6 +180,18 @@ static void fill_fspm_cpu_params(FSP_M_CONFIG *m_cfg,
...
+		m_cfg->HyperThreading = get_hyper_threading_option();
+
+	m_cfg->ActiveCoreCount = get_active_core_count_option();
+	m_cfg->ActiveSmallCoreCount = get_active_small_core_count_option();
+
+	if (m_cfg->ActiveCoreCount == 0 && m_cfg->ActiveSmallCoreCount == 0) {
```
SergiiDmytruk, line 190: You've mentioned in Jira that disabling all P-cores doesn't work, while this check suggests that it's possible to disable all of them.  So which is it?

miczyg1: It doesn't but there are SKUs that only have E-cores. And this check is purely added to avoid a soft-brick by injecting invalid values into EFIVAR store. Disabling all cores should never be possible, but variables are just an external input, which may be malicious/invalid.

src/soc/intel/alderlake/romstage/fsp_params.c
miczyg1: I have hardcoded the default to true in VFR, so I will have to update DMP side if we want to depend on CONFIG(FSP_HYPERTHREADING)

src/soc/intel/alderlake/romstage/fsp_params.c
miczyg1: I think we won't get around it without actually adding the SKU table to coreboot and checking the settings against the number of cores of given type...

src/soc/intel/alderlake/romstage/fsp_params.c
miczyg1: Updated DMP wit ha new PCD for HT default state. ALso moved the get_uint_option call to vc/dasharo

src/soc/intel/alderlake/romstage/fsp_params.c
miczyg1: Added the SKU table and refactored the checks

src/soc/intel/alderlake/Makefile.mk
```diff
@@ -27,6 +28,7 @@ romstage-y += reset.c
 ramstage-y += acpi.c
 ramstage-y += chip.c
 ramstage-y += cpu.c
+ramstage-y += cpu_util.c
```
SergiiDmytruk, line 31: I think `romstage` is enough.

payloads/external/edk2/Makefile
```diff
@@ -345,6 +345,22 @@ endif
...
+# PcdShowCpuCoreDisable = FALSE
+ifeq ($(CONFIG_EDK2_CORE_DISABLE_OPTION),y)
+BUILD_STR += --pcd gDasharoSystemFeaturesTokenSpaceGuid.PcdShowCpuCoreDisable=TRUE
+endif
+# PcdShowCpuHyperThreading = FALSE
+ifeq ($(CONFIG_EDK2_HYPERTHREADING_OPTION),y)
+BUILD_STR += --pcd gDasharoSystemFeaturesTokenSpaceGuid.PcdShowCpuHyperThreading=TRUE
+endif
+# PcdCpuHyperThreadingDefault = TRUE
+ifneq ($(CONFIG_EDK2_HYPERTHREADING_DEFAULT_STATE),y)
```
SergiiDmytruk, line 361: This breaks build for boards which use EDK without this option.  Need to either handle this option differently or update other boards with EDK to a new revision which could be done like:

```
sed -Ei '/Dasharo\/edk2(\.git)?"/,+1s/="[0-9a-fA-F]{40}"/="a7b33b10e9d86bf68e6e8cb89cd910fb6bc370c9"/g' configs/config.*
```

SergiiDmytruk (changes requested):

src/soc/intel/alderlake/Makefile.mk
miczyg1: For now yes.

https://github.com/Dasharo/coreboot/pull/518/commits/62fadd264bb206231bd51921bf4cc1a7e1731a3d#diff-3cbb42793ae5edb247eb7ed23b831be01a29436e0c036a68eb83460c6991941aR21

payloads/external/edk2/Makefile
miczyg1: Thanks, applied

miczyg1: 236a12e978f80764ae9f86c5f8a1ae859ae81726

SergiiDmytruk (approved):

PR #528, opened 2024-06-25
src/soc/intel/alderlake/Kconfig: Disable RAMTOP caching

miczyg1:
Caching the RAMTOP causes issues with FSP MemoryInit for platforms with DIMM modules. When memories are swapped, the training may sometimes hang randomly. Seems to be a manifestation of the issue described here: https://review.coreboot.org/c/coreboot/+/81269/ (despite the L3 cache sets is power of 2 on the tested i5-1235U SKU).

Disabling the RAMTOP caching make the memory training reliable no matter how many times the DIMM modules are swapped/changed in slots.


src/soc/intel/alderlake/Kconfig
```diff
@@ -40,7 +40,7 @@ config SOC_INTEL_ALDERLAKE
 	select CPU_INTEL_COMMON_VOLTAGE
 	select SOC_INTEL_COMMON_ACPI_WAKE_SOURCE
 	select SOC_INTEL_COMMON_BASECODE
-	select SOC_INTEL_COMMON_BASECODE_RAMTOP
+	# select SOC_INTEL_COMMON_BASECODE_RAMTOP
```
macpijan, line 43: It should be disabled on SoC level for all ADL boards, not on the mainboard level?
How about upstreaming such a change?

miczyg1: I doubt it will get merged. This "feature" is mostly for boot time optimization for Chromebooks.

If it causes problem on one board, then it will cause also problems on other boards as well most likely. We just haven't exercised memory compatibility that hard on MSI or ADL laptops.

macpijan: So enable for CHROME_OS only perhaps?
There must be a way for keeping chromebooks happy and not breaking other boards at the same time, that is acceptable for upstream.

miczyg1: Ok, guarded with CHROMEOS

krystian-hebel (commented):
For commit message, I'd rather link to https://review.coreboot.org/c/coreboot/+/78091, it is closer related to our issues.

It may be worth to upstream it ASAP and get feedback from the community.

krystian-hebel:
> > For commit message, I'd rather link to https://review.coreboot.org/c/coreboot/+/78091, it is closer related to our issues.
> 
> Actually not, it is nowhere close to what we observe. We don't even end up with FSP MemoryInit error printed in coreboot, because we don't return from FSP...

And runtime hangs are closer?

miczyg1:
> > > For commit message, I'd rather link to https://review.coreboot.org/c/coreboot/+/78091, it is closer related to our issues.
> > 
> > 
> > Actually not, it is nowhere close to what we observe. We don't even end up with FSP MemoryInit error printed in coreboot, because we don't return from FSP...
> 
> And runtime hangs are closer?

Yes, given that RAMTOP caching sets additional MTRR for romstage+postcar only. In ramstage the MTRRs are set to final values, and RAMTOP MTRR gets replaced by a WB MTRR covering whole RAM memory under 4G.

krystian-hebel (approved):

PR #529, opened 2024-07-01
security/vboot/Makefile.mk: silence warnings when signing binary

mkopec:
futility always prints a warning when building:

WARNING: prepare_slot: VBLOCK_A keyblock is invalid.

This is the expected state, because at that point vblock hasn't been created yet.

Pipe warnings to /dev/null to prevent worrying users unnecesarily.

Change-Id: I93fd5dafe220bafcf5325ecd4f9f4ee94fc3f81f

### DIFF
--- a/src/security/vboot/Makefile.mk
+++ b/src/security/vboot/Makefile.mk
@@ -387,7 +387,7 @@ files_added:: $(obj)/coreboot.rom $(FUTILITY) $(CBFSTOOL)
 		--kernelkey "$(CONFIG_VBOOT_KERNEL_KEY)" \
 		--version $(CONFIG_VBOOT_KEYBLOCK_VERSION) \
 		--flags $(CONFIG_VBOOT_KEYBLOCK_PREAMBLE_FLAGS) \
-		$(obj)/coreboot.rom
+		$(obj)/coreboot.rom 2> /dev/null
 	if [ "$(CONFIG_VBOOT_SLOTS_RW_AB)" = 'y' ]; then \
 		printf "    FLASHMAP Layout generated for RO, A and B partition.\n"; \
 	elif [ "$(CONFIG_VBOOT_SLOTS_RW_A)" = 'y' ]; then \


macpijan:
```
futility always prints a warning when building:

WARNING: prepare_slot: VBLOCK_A keyblock is invalid.

This is the expected state, because at that point vblock hasn't been created yet.
```

This sounds odd, is it something that should/could be fixed in the tool?
Why tool prints a warning, if the state is as expected?

I am a bit worried that this solution will get rid of all other messages as well, not only this case.

macpijan:
> I would change the loglevel to info and maybe reword the message to be less alarming. We'd have to send a contribution to upstream vboot and wait until it's merged and then the submodule gets bumped in coreboot.

To get rid of this in the long run, we need to do this anyway, so why not right away? And in such scenario, I think a workaround as proposed in this MR is also fine, as long as it is linked to the upstream contribution. To avoid the complexity of fork maintenance.

miczyg1:
IIRC the `prepare_slot` is used not only during the VLBOKC creation and signing but also for printing and validation. That's the problem. Whatever you do, VBLOCK will be checked and the warning be printed. Question is how to approach it? Detect which operation is being processed? Skip the warning on signing and creation?

macpijan:
I do not know the details here, but on the higher level it sounds to me similar to such case: https://review.coreboot.org/c/coreboot/+/82692

miczyg1:
> I wanted to just append `Will create new keyblock.` to the current warning. That would hopefully be less alarming than simply `keyblock is invalid.`

Sounds good, if it is sufficient...

mkopec:
Patch changing loglevel and expanding message sent upstream: https://chromium-review.googlesource.com/c/chromiumos/platform/vboot_reference/+/5682443

macpijan:
@mkopec Can you please add it to the commit msg? Then feel free to merge.

macpijan (approved):

mkopec:
@macpijan added upstream reference to commit message

PR #530, opened 2024-07-03
Wmowinski dasharo onboarding

wiktormowinski:


macpijan:
No need for MR here, these changes should be simply documented under https://github.com/Dasharo/presentations/blob/main/archive/dasharo-onboarding.md

PR #531, opened 2024-07-05
Add deploy workflow

PLangowski:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,241 @@
...
+          base_dir="protectli"
+          model=$(echo "$tag" | cut -d'_' -f1-3)
+          release=$(echo "$tag" | cut -d'_' -f4)
+          echo "base_dir=$base_dir" >> "$GITHUB_OUTPUT"
+          echo "model=$model" >> "$GITHUB_OUTPUT"
+          echo "release=$release" >> "$GITHUB_OUTPUT"
+
+      # Allows downloading artifacts from a different workflow
+      - name: Download workflow artifact
+        uses: dawidd6/action-download-artifact@v6
```
macpijan, line 39: Why do we need to use action from some user, not from GH: https://github.com/actions/download-artifact ?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,241 @@
...
+          workflow: build.yml
+          name: ${{ steps.artifact_name.outputs.artifact_name }}
+          path: "./artifacts"
+
+      - name: Upload to Nextcloud
+        env:
+          CLOUD_URL: ${{ secrets.CLOUD_URL }}
+          CLOUD_PASSWORD: ${{ secrets.CLOUD_PASSWORD }}
+        run: |
+          BASE_URL="https://cloud.3mdeb.com/public.php/webdav"
```
macpijan, line 228: It's a bit painful to see the amount of code duplication here.

When pasting this yaml to chatgpt, it suggests to use: https://docs.github.com/en/actions/using-workflows/reusing-workflows

That would be ideal. 

If not this, perhaps at least putting some of these duplicated commands into script would work.

macpijan (changes requested):

.github/workflows/deploy.yml
PLangowski: The "official" action does not allow downloading artifacts from a different workflow.

PR #535, opened 2024-07-17
mb/protectli/vault_jsl/mainboard.c: disable ASPM L0s for WiFi

krystian-hebel:
When L0s is enabled for WiFi card, a lot of errors are reported in dmesg, both for Data Link Layer and Physical Layer. L1 doesn't show these problems, so it is left enabled. The same behavior is present with AMI firmware, so it is believed to be a hardware issue.

Change-Id: Id3f9fec5f99b1eead197450f42ee78241be77393


krystian-hebel:
Fixes https://github.com/Dasharo/dasharo-issues/issues/943

src/mainboard/protectli/vault_jsl/mainboard.c
```diff
@@ -31,6 +31,9 @@ void mainboard_silicon_init_params(FSP_S_CONFIG *params)
 	params->PcieRpLtrEnable[5] = 1;
 	params->PcieRpLtrEnable[6] = 1;
 
+	/* Enable L1 only for WiFi, L0s doesn't work */
+	params->PcieRpAspm[4] = 2;
```
miczyg1, line 35:
```suggestion
	/*
	 * Atheros WiFi cards don't like ASPM L0s, disable it.
	 * ASMedia Switch on V1610 supports only ASPM L1, so no-op */
	if (!CONFIG(BOARD_PROTECTLI_V1610))
		params->PcieRpAspm[4] = FSP_PCH_PCIE_ASPM_L1;
```

And add `#define FSP_PCH_PCIE_ASPM_L1 2` somewhere.

miczyg1 (changes requested):
I would also test it on V1610, because the WiFi is behind a switch.

src/mainboard/protectli/vault_jsl/mainboard.c
miczyg1: @krystian-hebel ping

krystian-hebel: Added: https://github.com/Dasharo/coreboot/compare/20b32be69f8af07d3079c61c1a81cd752c64f156..1c135d6ca4356ea449810e8224955b62b73a070c

PR #538, opened 2024-07-23
VP6670 ESXi fix

miczyg1:
MADT LAPIC entries were generated for all CPUs in range 0 to CONFIG_MAX_CPUS regardless of their existence. This caused the cpu->path.apic.core_type field to have a different value than specified in cpu_perf_eff_type enum and extra efficient cores being added to the list with an unusuallly high APIC ID dereferenced from NULL pointer. For systems which obtain LAPICs/CPUs topology from MADT, like ESXI, it resulted in finding more CPUs than available in the system (exactly one more core in case of ESXi, because other entries had the same ACPI ID and were ignored as duplicates). SMP initialization failed in such case with timeout on waking a non-existent CPU core.

TEST=Boot ESXi in headless mode on VP6670.


krystian-hebel (approved):

macpijan:
Should be applicable for upstream as well?

PR #544, opened 2024-07-26
configs/config.novacustom_v5*: Add gbe blob paths

philipandag:


mkopec (commented):
We also need to bump the dasharo-blobs submodule right?

philipandag:
I have been struggling with it but it should be ok now, I think

mkopec (approved):

PR #546, opened 2024-08-08
Disable TPM2 PCR banks which aren't used by coreboot

SergiiDmytruk:
coreboot can only extend a single PCR bank of TPM2 and this change results in all available PCRs being extended.

This is implemented by passing hash mask to EDK2 because coreboot would need to implement `TPM2_PCR_Allocate` command to select PCR banks on its own while EDK already has it.  We're also primarily interested in solving this for EDK2 case anyway.

This affects all Dasharo EDK2 boards, but can be limited to selected ones if some negative side-effects are expected.

See https://github.com/Dasharo/dasharo-issues/issues/982 for context.


miczyg1:
So, we expect to have the correct PCR banks allocated after a reboot (after EDK2 disables the banks we don't want to have) ?

SergiiDmytruk:
There is a cold reboot to reconfigure TPM which EDK does automatically if active PCR banks don't match build-time configuration. The user should never see unsupported PCRs being enabled, if that's what you're asking.

payloads/external/edk2/Makefile
```diff
@@ -373,6 +373,20 @@ ifneq ($(CONFIG_CPU_MAX_TEMPERATURE),)
 BUILD_STR += --pcd gDasharoSystemFeaturesTokenSpaceGuid.PcdCpuMaxTemperature=$(CONFIG_CPU_MAX_TEMPERATURE)
 endif
 endif
+# gEfiSecurityPkgTokenSpaceGuid.PcdTpm2HashMask = 0x0000001f
+# The tested options are mutually exclusive (there is no option for SM3 256).
```
krystian-hebel, line 377: Could you elaborate in the comment that this is design decision made by coreboot, preferably with link to either change that caused it or coreboot (not Dasharo) issue with discussion?

This WILL impact TrenchBoot...

SergiiDmytruk: > Could you elaborate in the comment that this is design decision made by coreboot, preferably with link to either change that caused it or coreboot (not Dasharo) issue with discussion?

https://github.com/Dasharo/coreboot/compare/8a68c4eb9a6e7bc00f9234f6614b8220bf32038e..1cb3194d58b43e2c36d440992fec2914df3e67df

> This WILL impact TrenchBoot...

https://github.com/Dasharo/dasharo-issues/issues/982#issuecomment-2289055801

krystian-hebel (approved):
Approved, https://github.com/Dasharo/edk2/pull/139 must be merged before this can land on `dasharo`.

PR #553, opened 2024-08-28
soc/intel/meteorlake: Hook the VT-d DMA protection option

philipandag:
Applied Miczyg1's patch to coreboot alder lake https://review.coreboot.org/c/coreboot/+/68450 to meteor lake. It should hopefully fix this issue https://github.com/Dasharo/dasharo-issues/issues/985 but I can't test it now because of missing FSP.

### DIFF
--- a/src/soc/intel/meteorlake/romstage/romstage.c
+++ b/src/soc/intel/meteorlake/romstage/romstage.c
@@ -156,4 +156,7 @@ void mainboard_romstage_entry(void)
 	pmc_set_disb();
 	if (!s3wake)
 		save_dimm_info();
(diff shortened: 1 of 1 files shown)


philipandag:
After applying the patch `cbem -l` says:
```
[ERROR]  VT-d PMR HOB not found, not enabling DMA protection
```

miczyg1:
The change is correct, however if the FSP does not produce the HOB, then there is not much we can do about it...

mkopec (approved):
Change looks good, actual DMA protection will work once it's fixed in FSP

PR #555, opened 2024-09-04
mb/protectli/vault_adl_p/vboot-rwa.fmd: Remove the hole in RW_MISC

miczyg1:
The hole in FMAP layout caused the Openness Score to fail counting the regions sizes, which didn;t sump up to image size. Enlarge the BOOTSPLASH region size to leave the RW_A and WP_RO without changes.


philipandag (approved):
The total size of the regions stays the same so I thinks it's OK.

PR #566, opened 2024-09-30
Mtl-h update FSP 4122_21

philipandag:


miczyg1:
@mkopec do we still need this PR? It looks like it uses `Dasharo/novacustom-blobs`

mkopec:
we don't, closing

PR #564, opened 2024-09-24
.github/workflows/deploy-template.yml: Add deployment for ODROID H4

miczyg1:

PR #573, opened 2024-10-15
Hwio 2024

pietrushnic:
We still have to decide which fixes/improvements to this branch will be merged with Dasharo.


miczyg1:
Besides the BTG config there is nothing to merge. The `build.sh` support still hangs on https://github.com/Dasharo/coreboot/pull/564 which still hasn't been fixed yet...

pietrushnic:
@miczyg1 I don't know if it should be merged as stated in the description of PR. My key question is probably regarding switching to dasharo-sdk instead of using coreboot-sdk here and dasharo-sdk in other places. For now, it could be just for Odroid, and then it could follow with different platforms.

PR #577, opened 2024-10-23
ec/system76/ec/system76_ec.c: Rework EC FW sync

miczyg1:


miczyg1:
@mkopec it seems to work now. Made a few update attempts in a row and none failed with AAI programming. Though I tested only a couple unused sectors... When I will be at office, I can test full flash update, but to be safe I would like to have a EC kit at hand. You may already test it on your end. No EC FW changes introduced.

mkopec:
I've been flashing back & forth between v0.9.0-rc10 and final with no issues so far :+1:

src/lib/hexdump.c
```diff
@@ -33,18 +33,18 @@ void hexdump(const void *memory, size_t length)
...
+				printk(BIOS_INFO, "   ");
+			printk(BIOS_INFO, "  ");
 			for (j = 0; j < num_bytes; j++)
-				printk(BIOS_DEBUG, "%c",
+				printk(BIOS_INFO, "%c",
 				       isprint(line[j]) ? line[j] : '.');
-			printk(BIOS_DEBUG, "\n");
+			printk(BIOS_INFO, "\n");
 		} else if ((all_zero == 2) || (all_one == 2)) {
-			printk(BIOS_DEBUG, "...\n");
```
mkopec: is this meant to be in this change?

mkopec:
I also think we may want to load the EC update file uncodnitionally, to make the TPM measurements consistent between AC and DC, but that doesn't need to go in this PR

src/lib/hexdump.c
miczyg1: Certainly not.. Used it to lower the debug level

miczyg1: Removed after rebasing.

mkopec (dismissed):

mkopec (approved):

PR #579, opened 2024-10-28
Refactor build.sh to be usable as the standard way of building Dasharo coreboot

philipanda:


build.sh
```diff
@@ -9,215 +9,202 @@ usage() {
...
+
+  echo -e "\tv540tu         - build Dasharo for Novacustom V540TU"
+  echo -e "\tv560tu         - build Dasharo for Novacustom V560TU"
+  echo -e "\tns5x7x         - build Dasharo for Novacustom NS5x/NS7x"
+  echo -e "\tnv4x_adl       - build Dasharo for Novacustom NV4x 12 gen"
+  echo -e "\tnx5x7x         - build Dasharo for Novacustom Nx5x/Nx7x"
+  echo -e "\tnv4x_tgl       - build Dasharo for Novacustom NV4x 11 gen"
+
+  echo "Used environmental variables:"
+  echo -e "\tCB_REVISION          - build Dasharo coreboot from a given revision name"
```
mkopec, line 39: I'm not sure if this should be in this script, if it only checks out a specific rev. For example what happens if it checks out to a revision that doesn't have this script? It may stay in memory and finish the run, but after it finishes the build.sh script may be different and not work the same way if someone calls it again

philipanda: That is true. The script is a part of the repository which it would checkout to a different version itself. I wonder if using the script as a universal way of building any version of Dasharo would be possible without making changes in historic revisions...

philipanda: Maybe it would be possible to get the ID of the current commit `git rev-parse HEAD` and restore it after building? This sounds like a cheap solution though...

philipanda:
Changed to draft as the refactoring changes were not thoroughly checked and the objective, to make it possible to also build FW for all the NC laptops was not yet achieved

PR #580, opened 2024-10-31
security/tpm: fix tis_drivers for postcar

ansiwen:
postcar was missing in the list of stages that need tis_driver initialization.

This is security critical, because postcar can not measure the ramstage into the PCR, at least if the TPM is already initialized in the bootblock or verstage. This means the measurement chain is interrupted and the PCR values are „broken“, because the ramstage can be modified without changing the PCR value.

### DIFF
--- a/src/security/tpm/tis.h
+++ b/src/security/tpm/tis.h
@@ -100,7 +100,7 @@ static inline bool tpm_first_access_this_boot(void)
 	(CONFIG(VBOOT_SEPARATE_VERSTAGE) && ENV_SEPARATE_VERSTAGE) ||  \
 	(CONFIG(VBOOT_STARTS_IN_ROMSTAGE) && ENV_RAMINIT) ||           \
 	(CONFIG(TPM_MEASURED_BOOT_INIT_BOOTBLOCK) && ENV_BOOTBLOCK) || \
-	(CONFIG(TPM_MEASURED_BOOT) && (ENV_RAMINIT || ENV_RAMSTAGE))
+	(CONFIG(TPM_MEASURED_BOOT) && (ENV_RAMINIT || ENV_POSTCAR || ENV_RAMSTAGE))
 #define __tis_driver __attribute__((used, section(".rodata.tis_driver")))
 #else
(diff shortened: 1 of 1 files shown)


pietrushnic:
@ansiwen, thanks for your contribution.

SergiiDmytruk:
Has this caused a build error with some configuration?  Looks like it could, but just wondering why it wasn't noticed.

By the way, this macro will be gone after the next rebase because upstream has an updated version of the commit where it's implemented differently.

ansiwen:
> Has this caused a build error with some configuration? Looks like it could, but just wondering why it wasn't noticed.

Sorry, I should have described the issue better. It builds fine but results in insecure PCR values. I updated the description now. 

> By the way, this macro will be gone after the next rebase because upstream has an updated version of the commit where it's implemented differently.

Yes, I know, this issue is downstream only.

ansiwen:
For what it’s worth, I added a bail-out for these cases in our branch to be on the safe side: https://github.com/Nitrokey/nethsm-coreboot/commit/38cd5a6fa95257a079330c164917f456459e1c59

SergiiDmytruk (approved):
> It builds fine but results in insecure PCR values.

Right, the driver just isn't on the list when it should be.  That's much worse than a build issue.  Thank you!

CI seems  to fail due to s missing token for non-public blobs, the code is fine.

SergiiDmytruk:
Oh, signing is required. @ansiwen, can you GPG-sign the commit or should I just amend and sign it myself?

ansiwen:
> Oh, signing is required. @ansiwen, can you GPG-sign the commit or should I just amend and sign it myself?

I never set up signing so far. Can I sign with Github? Otherwise it's easiest if you amend it.

ansiwen:
Oh, I should be able to sign with my ssh key, let me try that...

ansiwen:
@SergiiDmytruk done.

SergiiDmytruk:
> Oh, I should be able to sign with my ssh key, let me try that...

Nice, I didn't know that will also work.

PR #581, opened 2024-11-05
mb/msi/ms7d25,7eo6: Fix USB 3.x port mapping

miczyg1:
The USB-C port is Gen2x2 and occupies two USB3.x line pairs. The missing second pair caused all the ports to be shifted by one and port 10 not being enabled as USB3.x capable. As a result one of the JUSB3 ports was working in HighSpeed only.

### DIFF
--- a/src/mainboard/msi/ms7d25/devicetree.cb
+++ b/src/mainboard/msi/ms7d25/devicetree.cb
@@ -57,33 +57,33 @@ chip soc/intel/alderlake
 					 "M2_1" "SlotDataBusWidth4X"
 		end
 		device ref xhci on
-			register "usb2_ports[0]" = "USB2_PORT_SHORT(OC2)" # USB-C LAN_USB1
-			register "usb2_ports[1]" = "USB2_PORT_SHORT(OC1)" # MSI MYSTIC LIGHT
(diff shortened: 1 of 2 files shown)


pietrushnic:
@miczyg1, how can we prove or validate that this patch fixes the issue? I'm completely fine merging based on any tests, your test posted here, my test if I understand how to do that, or maybe even a community test if there is an issue associated.

miczyg1:
> @miczyg1, how can we prove or validate that this patch fixes the issue? I'm completely fine merging based on any tests, your test posted here, my test if I understand how to do that, or maybe even a community test if there is an issue associated.

The issue has been reported on matrix. The test is simple: USB 3.x device should be visible in the system with USB 3.x speed in any USB 3.x capable port.

The main cause of the changes was the USB 3.x port 10 present on JUSB3, associated with USB 2.0 port 10. When a USB 3.x device was connected to this port, it always appeared as USB 2.0 device (in USB2.0 port 10) instead of USB 3.x device (in USB 3.x port 10).

I have verified the changes using Windows USB Device Tree Viewer. Can also be verified using lsusb on Linux (the device would appear under 3.0 root hub instead of 2.0 root hub).

PR #597, opened 2024-12-03
Protectli vault adl n typec

miczyg1:


miczyg1:
Tested on Protectl iVP2410 and VP2410.

1. VP2410 previously didn't work (enable/disable SHA1 PCR bank). Now it works.
2. VP2420 previously worked (enable/disable SHA1 PCR bank) and still works.

The change causes the coreboot's TPM PPI buffer in ACPI not being used. However we have no tests to verify if TPM PPI works in OS currently... Switching to the PPI from SecurityPkg allows to get the TPM2 menu features working properly in firmware.

mkopec:
@miczyg1 I'm not sure I see the relation between these changes and TPM PPI, is that a mistake?

miczyg1:
> @miczyg1 I'm not sure I see the relation between these changes and TPM PPI, is that a mistake?

Wrong PR...

miczyg1:
Sorry, it is not meant for review yet. Uploaded this PR only for testing purposes...

miczyg1:
Now it is ready for review. I looks like I forgot about PMC mux conn entries in devicetree, but they caused the Type-C display to not work anymore. As a w/a I have instructed the FSP to program AUX bias pads. Details in commit messages.

PR #600, opened 2024-12-09
Nvc usb charge port

miczyg1:


src/vendorcode/dasharo/include/dasharo/options.h
```diff
@@ -36,6 +36,13 @@ enum cse_disable_mode {
 	ME_MODE_DISABLE_HMRFPO = 10
 };
 
+enum usb_port_power_opt {
+	USB_PORT_ON_WHEN_POWERED = 0,
+	USB_PORT_ALWAYS_ON = 1,
+	USB_PORT_WHEN_ON_AC = 2,
```
mkopec:
```suggestion
	USB_PORT_ON_WHEN_ON_AC = 2,
```

I think that makes more sense

PR #602, opened 2024-12-11
[NOT FOR MERGE] src/soc/intel/common/block/cpu/Kconfig: Disable TME by default

miczyg1:

PR #603, opened 2024-12-11
[DNM] NovaCustom: Disable TME

mkopec:


mkopec:
ah well https://github.com/Dasharo/coreboot/pull/602 is better, closing

PR #607, opened 2025-01-03
config.hardkernel_odroid_h4: change EDK2 revision

m-iwanicki:
Fixes missing subhook repository and failed git clone


pietrushnic:
@m-iwanicki, what tests did you do to confirm this change works?

m-iwanicki:
Just tested that it builds (it didn't before because of missing subhook repository).
Commit used is identical to one already on dasharo branch: https://github.com/Dasharo/coreboot/blob/6233d0e15229d0cfd070fc97cb407c2d5767a052/configs/config.hardkernel_odroid_h4#L29

pietrushnic:
Yeah, I'm unsure if that is enough to confirm that the exercises will look the same and we will have no regression.

miczyg1:
> Yeah, I'm unsure if that is enough to confirm that the exercises will look the same and we will have no regression.

Well you have to merge it anyway, otherwise it will not build, because a public submodule used by edk2 has gone private...

mkopec:
here's my workaround for the missing subhook submodule that doesn't require touching the code: https://github.com/Dasharo/fwbuild/blob/237d95870ee3a79617165a109a18e3a3f426254f/v540tu/uefi/v0.9.0/bios/FWBUILD#L64-L65

macpijan:
Also, there was a branch before set in the config, which is not reproducible. So we do not really know for sure, from which revision these binaries have been built before.

If we need to be sure that there won't be regressions, we can use the very same binaries we have used before - if we still have them somewhere.

Going further, why do we need the hwio branch? It would be best if we could simply use one of the tagged releases (even RC tags) for the training purposes.

I see no other way in moving forward here as merging this PR. In the next step, we can consider whether we need a separate branch, as using a tag or even a specific commit from `dasharo` branch would help to avoid desync.

PR #611, opened 2025-02-06
WIP: Spd dache enabled

miczyg1:


miczyg1:
This PR is obsolete, it intended to implement SPD caching which was not the solution to the problem. The fix for the problem is here: https://github.com/Dasharo/coreboot/pull/618

PR #612, opened 2025-02-07
Mtl igpu pmax

mkopec:


src/mainboard/clevo/mtl-h/variants/igpu/overridetree.cb
```diff
@@ -1,4 +1,16 @@
 chip soc/intel/meteorlake
+	# Power limits,
+	# PsysPL2, PsysPL3, PL4 are configured by EC at runtime
+	register "power_limits_config[MTL_P_682_482_CORE]" = "{
+		.tdp_pl1_override = 35,
+		.tdp_pl2_override = 64,
+		.tdp_pl4 = 120,
```
miczyg1: Not sure if pl4 can be bigger than pmax. I know EC will limit it, but I don't know what the consequences may be.

How much Watts does the power supply have?

mkopec: PSU is 90W, but Pmax has to be 99W because that's the scale that the battery charger uses to report Psys to CPU.

I thought PL4 can be up to PSU rating + max battery rating but it feels like every doc I read about it says something different. Will adjust PL4 to the same value as PMAX to be on the safe side.

mkopec: https://github.com/Dasharo/coreboot/commit/ed3fdaec053cf201ed16a54a564b7455416b9bdf#diff-5440c50c876ba495f3ff1099c44d020892a1941250b8987c8aecea3d3e47d80fR7

mkopec:
Fixes https://github.com/Dasharo/dasharo-issues/issues/1216

src/mainboard/clevo/mtl-h/variants/igpu/overridetree.cb
miczyg1: @mkopec as I suspected. PsysPmax = PL4 + ROP MAX. See doc 634413 on our cloud (section 3.4). Basically we need a margin for Psys PMAX equivalent to Rest of Platform (ROP) power here.

> that's the scale that the battery charger uses to report Psys to CPU.

Scale or granularity? Where can I check it?

mkopec: Scale, as determined by Rpsys resistor. See schematics page B-40 `MP29001 Controller`

So I guess we should reduce PL4 by ROP max, need to determine it somehow first

miczyg1: Fair enough. But if to believe what the doc says, PL4 should be lower than Psys PMAX. To be on the safe side, I would give PL4=90W maybe?

mkopec: Done https://github.com/Dasharo/coreboot/pull/612/commits/86233ac56e4c18e398efd9b505924e29aeb9d029

miczyg1 (approved):

miczyg1:
@mkopec needs rebase to fast-forward merge it

mkopec:
rebased, let's see what CI thinks

PR #613, opened 2025-02-07
Protectli no shell and no ipxe timeout

miczyg1:


macpijan (dismissed):
lgtm provided that we resolve https://github.com/Dasharo/edk2/pull/208/files first

PR #614, opened 2025-02-07
conform.yaml: update to coreboot guidelines

macpijan:
Change first line of commit message from 80 to 75.

Set the commit body as required.

Source:
https://doc.coreboot.org/contributing/gerrit_guidelines.html#commit-message-guidelines

Change-Id: Iaab7a6a2c4adffa6b04b170b94cc98f3b157558e

### DIFF
--- a/.conform.yaml
+++ b/.conform.yaml
@@ -3,11 +3,11 @@ policies:
   - type: commit
     spec:
       header:
-        length: 80
+        length: 75
         imperative: false
         invalidLastCharacters: .
       body:
-        required: false
+        required: true
       dco: true
       gpg:
         required: true


macpijan:
Testing if it works as expected.

This conform tool should allow us to check if commit body exists.

For the Upstream-Status tag we would likely need a separate hook.

.conform.yaml
```diff
@@ -3,11 +3,11 @@ policies:
   - type: commit
     spec:
       header:
-        length: 80
+        length: 75
         imperative: false
         invalidLastCharacters: .
       body:
```
krystian-hebel, line 9: Can the length be specified separately for body? There is also a limit for it, and it used to be 72 (see e.g. https://review.coreboot.org/c/coreboot/+/82041/8..9//COMMIT_MSG#b12), I don't know if it changed recently or not.

macpijan:
Ok, the body check does not work for me, at least the way I would expect. Need to check.

EDIT:

OK I guess it's due to the Gerrit hooks adding: `Change-Id: Iaab7a6a2c4adffa6b04b170b94cc98f3b157558e`.

So we would need to modify this tool to add some more exceptions, preferably configurable via config file. Currently, it drops the DCO (`Signed-off`) line from this check only.

.conform.yaml
krystian-hebel: Seems to be new: https://review.coreboot.org/c/coreboot/+/85915

macpijan: At present, the body is just checked if it exists. There is no enforcement on the line length whatsoever, it seems. So this would also need to be implemented.

PR #625, opened 2025-02-26
src/mainboard/clevo/mtl-h/variants/dgpu/gpio.c: disable dgpu initiali…

wiktormowinski:
…zation


mkopec:
This was done as a test for power consumption, I guess it can be closed now.

PR #627, opened 2025-03-05
Vp2430 new rev

miczyg1:
TEST=Boot Ubuntu 22.04 live on the new VP2430 revision. All displays work, tested HDMI, DP and both USB-C ports with Dell monitors.


src/soc/intel/common/block/graphics/graphics.c
```diff
@@ -455,6 +455,8 @@ static const unsigned short pci_device_ids[] = {
 	PCI_DID_INTEL_ADL_N_GT1,
 	PCI_DID_INTEL_ADL_N_GT2,
 	PCI_DID_INTEL_ADL_N_GT3,
+	PCI_DID_INTEL_ADL_N_GT4,
```
macpijan, line 458: Is it part of upstream already?

If yes, we should probably cherry-pick and indicate that.

If not, is it worth to send it right away, and point to the upstream review via `Upstream-Status` tag?

macpijan (approved):
my note is not blocking

src/soc/intel/common/block/graphics/graphics.c
miczyg1: It is not there yet.

miczyg1: https://review.coreboot.org/c/coreboot/+/86750

miczyg1: Amended the commit: https://github.com/Dasharo/coreboot/pull/627/commits/73dbae643bf84372df90643304fd0a64c1c781b8

mkopec (approved):

PR #629, opened 2025-03-08
Add support to disable DGPU in EDK2

XutaxKamay:
I refer to this PR: https://github.com/Dasharo/edk2/pull/226

I've only tested NV4x with TGL & 1650 GTX since I have only this laptop but I believe it could be enabled for others laptops.
I get some power saving with this and it is currently a workaround for me to make s0ix working when ME is enabled (because nvidia driver doesn't work properly).


filipleple:
@XutaxKamay Thank you very much for your contribution! We are going to test the functionality, extend it to other laptops and finally ship it in future releases.

PR #634, opened 2025-03-23
RFC: Integrate CFR into Dasharo options

mkopec:
Add some generic Dasharo options and some example mainboard-specific options.

Not sure how to best handle options that only affect the payload. Might want to move them to coreboot to have a single centralized interface for enabling and hiding options

I'd also consider removing some code from dasharo vendorcode and using get_uint_option where possible, to reduce unupstreamable code to a minimum

Needs https://github.com/Dasharo/edk2/pull/230


mkopec:
Current state:

image

Mainboard specific options are defined in mainboard code. Common Dasharo options are defined in a shared header and included as needed in mainboard code. Seems to be working pretty well

src/vendorcode/dasharo/include/dasharo/options.h
```diff
@@ -63,6 +63,9 @@ struct battery_config {
 	uint8_t stop_threshold;
 } __packed;
 
+/* Initialize Dasharo CFR backend */
+void cfr_init_dasharo(void);
```
mkopec: unnecessary, remove

src/mainboard/lenovo/m900_tiny/cfr.c
```diff
@@ -0,0 +1,86 @@
...
+	&security,
+	&processor,
+	&chipset,
+	&devices,
+	NULL
+};
+
+void mb_cfr_setup_menu(struct lb_cfr *cfr_root)
+{
+	if (!CONFIG(BOOTMEDIA_SMM_BWP))
```
mkopec: Could be moved to dasharo/cfr.h and handled in a callback

PR #635, opened 2025-03-24
Dasharo driver

mkopec:
Supports temperatures, fan pwm and fan rpm on Dasharo EC platforms


SebastianCzapla:
On v540tnd, while working with .rom generated from this PR, suspend fails.
Command used:
```
fwts s3 -f -r /tmp/suspend_test_log.log
```
Results: Fans do not stop while suspended, after waking up, the keyboard is non-functional. SSH connection to woken up laptop fails with "No route to host" error. Fan turbo mode works.

PR #638, opened 2025-03-28
Cfr ng

miczyg1:


mkopec:
I've been using a thinkcentre tiny as my test mule. The current status is:

- All applicable options for the platform are added in cfr
- CFR option backend is hooked into dasharo vendorcode
- Dasharo menu is hidden, CFR menu is on the edk2 frontpage as Advanced Options
- The edk2 changes are a bit hacky: https://github.com/Dasharo/edk2/pull/232 but I managed to hook cfr variables into our boot policies
- I've sent cfr changes that are board specific for upstream review: https://review.coreboot.org/c/coreboot/+/87048

That's basically it. The CFR form is displayed as a single scrollable page, which is nice on a big monitor, but will wreak havoc on our tests over serial console. I didn't bother to call the options identically to our edk2 versions but it should be possible.

Overall it seems to work well, I haven't seen any major issues. Reset to default works fine even if you call it before entering the menu, which was a pain point with some previous dynamic form implementations.

pietrushnic:
@mkopec @miczyg1, can we describe the primary benefits to users, OEMs, and developers by introducing this change?

mkopec:
The benefit is that it will be easier to add options, especially board-specific options. Right now we have all options in the same place edk2 with PCDs to control their visibility, which adds a lot of code for each board specific option. CFR allows the mainboard to specify options for the payload. So options like these will be easier to add:
- dGPU disable
- SSD slot enable/disable
- [some other board specific device] enable/disable

plus hopefully it makes us closer to upstream

pietrushnic:
@mkopec ok, make sense. I wonder how much time we spent on those operations in the past and how much we will have to spent now. This would prove if it improves situation. For now we just have hipothesis.

Easier control of board specific options, as well as those who customer want/may have according to their plan, is important feature. Hopefully we can leverage that more efficiently when it would be merged.

PR #640, opened 2025-03-30
configs/config.novacustom_*: bump versions after rebase

mkopec:
Change-Id: Ie7a55f50826f4daa45260a327bc76437e236b297


mkopec:
GitHub GC deleted the old edk2 rev, needed to bump it for all platforms

miczyg1:
@mkopec edk2 bump can be dropped. Already on dasharo branch: https://github.com/Dasharo/coreboot/commit/9d8eec98859ba9a62c7077e45f9bd9a529667115

mkopec:
@miczyg1 ok, dropped

configs/config.novacustom_v540tnx
```diff
@@ -1,4 +1,4 @@
-CONFIG_LOCALVERSION="v0.9.1"
+CONFIG_LOCALVERSION="v1.0.0-rc1"
```
miczyg1: Why v1.0.0? We do not use v1.0.0 which would indicate production-ready FW.

miczyg1 (changes requested):

configs/config.novacustom_v540tnx
mkopec: The alternative was v0.10.0 which I expect would break a lot of version number parsers, or v0.9.3 which does not reflect that there was a major change in coreboot base

mkopec: @miczyg1 what do you think the version number should be?

miczyg1: v0.9.3

For MSI we did the same. We just bump the patch version regardless of base change

mkopec: @miczyg1 done in https://github.com/Dasharo/coreboot/pull/640/commits/9427f7a0709ca45a87c5e5911add2f9c104b9d71

miczyg1 (approved):

PR #642, opened 2025-04-01
Dgpu only rebased

mkopec:


mkopec:
@filipleple please take a look at this one as well. I've tested this on V540TNx and on NV4x TGL with the dGPU, seems to work well.

filipleple (approved):
@mkopec awesome, LGTM as well

PR #646, opened 2025-04-04
src/ec/dasharo/ec: Add support for Dasharo Linux driver

mkopec:
Change-Id: Ica3976097f87ef76bf795a5ce73f68bbf2f51c16

Third time's the charm


src/ec/dasharo/ec/acpi/dshr.asl
```diff
@@ -0,0 +1,147 @@
...
+#endif
+					Default { Return (0) }
+				}
+			}
+
+			Case (DASHARO_FEATURE_FAN_PWM)
+			{
+				Switch (ToInteger(Arg1))
+				{
+					Case (DASHARO_FAN_CPU) { Return (CONFIG_EC_DASHARO_EC_CPU_FAN_COUNT) }
```
miczyg1: Wouldn't this mean that if EC_DASHARO_EC_DGPU is enabled, then EC_DASHARO_EC_CPU_FAN_COUNT must be 1? Currently it is possible to misconfigure, e.g. EC_DASHARO_EC_DGPU=1 and EC_DASHARO_EC_CPU_FAN_COUNT=2

src/ec/dasharo/ec/Kconfig
```diff
@@ -29,3 +29,7 @@ config EC_DASHARO_EC_UPDATE_FILE
 config EC_DASHARO_EC_FLASH_SIZE
 	hex
 	default 0x20000
+
+config EC_DASHARO_EC_CPU_FAN_COUNT
+	int
+	default 1
```
miczyg1, line 47: Shouldn't be a range 1,2?

miczyg1 (changes requested):

src/ec/dasharo/ec/acpi/dshr.asl
mkopec: Done in https://github.com/Dasharo/coreboot/pull/646/commits/d7ffd78a91e4c2a3dc93eb78e238b2c381723c81

also fixed a divide by 0 error

src/ec/dasharo/ec/Kconfig
mkopec: Done in https://github.com/Dasharo/coreboot/pull/646/commits/d7ffd78a91e4c2a3dc93eb78e238b2c381723c81

src/ec/dasharo/ec/acpi/dshr.asl
```diff
@@ -0,0 +1,164 @@
...
+#if CONFIG(EC_DASHARO_EC_DGPU)
+					Case (DASHARO_FAN_GPU) { Return (1) }
+#endif
```
miczyg1:
```suggestion
#if CONFIG(EC_DASHARO_EC_DGPU)
					Case (DASHARO_FAN_CPU) { Return (1) }
					Case (DASHARO_FAN_GPU) { Return (1) }
#else
					Case (DASHARO_FAN_CPU) { Return (CONFIG_EC_DASHARO_EC_CPU_FAN_COUNT) }
#endif
```

Wouldn't it make more sense @mkopec ? We can't have more than 1 fan per component in dGPU case

mkopec: applied in https://github.com/Dasharo/coreboot/pull/646/commits/e68c19331c07be7c286c1adba5e8e821fcdcc59a

miczyg1 (approved):

PR #647, opened 2025-04-07
Vault adln usb wake

miczyg1:


miczyg1:
Won't work, the USB ports are not powered

PR #654, opened 2025-04-15
FTDI driver support

filipleple:


miczyg1 (dismissed):
Since you introduce a new Kconfig option for EDK2 you have to bump edk2 and edk2-platforms revisions in every config. Otherwise, this new option will remain broken on the platform which do not update the edk2 and edk2-platforms. @filipleple

filipleple:
Right, thanks, updated for all platforms

mkopec (approved):
I've tested a binary for NS5x TGL from CI and it seems to work as expected :+1:

PR #655, opened 2025-04-15
capsule_cabinet.sh: Add script for generating lvfs cabinets

mkopec:
Change-Id: Ic54ab3230362eafd3f002d0a45d7e44087c85393

### DIFF
--- a/capsule_cabinet.sh
+++ b/capsule_cabinet.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+# A script to generate cabinets for fwupd / LVFS
+
+set -e
+
+function die() {
+    echo error: "$@" 1>&2
+    exit 1
+}
+
+if [ $# -ne 1 ]; then
+    die "Incorrect number of input parameters specified: $# (expected: 1)"
+fi
+
+if [ -z $1 ]; then
+    die "No input capsule specified"
+fi
+
+if [ ! -f $1 ]; then
+    die "File $1 not found"
+fi
+
+if [ ! -f .config ]; then
+    die "No '.config' file in current directory"
+fi
+
+# import coreboot's config file replacing $(...) with ${...}
+while read -r line; do
+    if ! eval "$line"; then
+        die "failed to source '.config'"
+    fi
+done <<< "$(sed 's/\$(\([^)]\+\))/${\1}/g' .config)"
+
+if [ "$CONFIG_DRIVERS_EFI_UPDATE_CAPSULES" != y ]; then
+    die "Current board configuration lacks support of update capsules"
+fi
+
+capsule=$1
+date=$(stat -c %w $capsule | cut -d ' ' -f 1)
+vendor=$(cat .config | grep -e "CONFIG_VENDOR_.*=y" | cut -d '=' -f 1 | cut -d '_' -f 3- | awk '{ print tolower($0) }')
+version=$(echo $CONFIG_LOCALVERSION | tr -d 'v' | cut -d '-' -f 1)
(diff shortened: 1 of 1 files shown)


capsule_cabinet.sh
```diff
@@ -0,0 +1,85 @@
...
+    die "current board configuration lacks support of update capsules"
+fi
+
+if [ -z $1 ]; then
+    die "No input capsule specified"
+fi
+
+if [ ! -f $1 ]; then
+    die "File $1 not found"
+fi
```
SergiiDmytruk: Arguments should be checked first before looking for `.config` and sourcing it.  Can also add `$# -eq 1` check to not silently ignore extra arguments.

capsule_cabinet.sh
```diff
@@ -0,0 +1,85 @@
...
+  <summary>${CONFIG_MAINBOARD_SMBIOS_PRODUCT_NAME} ${CONFIG_MAINBOARD_VERSION} system firmware</summary>
+  <description>
+    <p>Dasharo ${CONFIG_MAINBOARD_SMBIOS_PRODUCT_NAME} ${CONFIG_MAINBOARD_VERSION} system firmware</p>
+  </description>
+  <provides>
+    <firmware type="flashed">${CONFIG_DRIVERS_EFI_MAIN_FW_GUID}</firmware>
+  </provides>
+  <url type="homepage">https://docs.dasharo.com/</url>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>LicenseRef-proprietary</project_license>
```
SergiiDmytruk, line 59: Proprietary Dasharo firmware?

capsule_cabinet.sh
```diff
@@ -0,0 +1,85 @@
...
+    <category>X-System</category>
+  </categories>
+  <custom>
+    <value key="LVFS::VersionFormat">quad</value>
+    <value key="LVFS::VersionFormat">dell-bios-msb</value>
+    <value key="LVFS::UpdateProtocol">org.uefi.capsule</value>
+  </custom>
+  <releases>
+    <release version="${version}" date="${date}" tag="${CONFIG_LOCALVERSION}" urgency="high">
+      <checksum filename="firmware.bin" target="content"/>
```
SergiiDmytruk, line 70: Since it's a capsule, maybe use `.cap` instead of `.bin`?

mkopec: fwupd requires that the firmware file inside the cabinet is named firmware.bin

capsule_cabinet.sh
mkopec: I didn't know what to put there, considering that the binaries also contain FSP, ME, and other blobs. I agree it's not a good look though

capsule_cabinet.sh
SergiiDmytruk: It seems to be the default name without the `<checksum>` tag: https://github.com/fwupd/fwupd/blob/15a1adfeb7c4010f38972fb0eb4ce2d2d040d9c6/src/fu-cabinet.c#L138-L141

A custom name is also used in the example metadata: https://lvfs.readthedocs.io/en/latest/metainfo.html

But we can keep it as is if there are compatibility concerns.

capsule_cabinet.sh
mkopec: Done in https://github.com/Dasharo/coreboot/pull/655/commits/db96eca47009cb121bb74745846bf1d554c32b5f

macpijan (approved):

PR #656, opened 2025-04-18
Vp2430 ESXi rc7

filipleple:


filipleple:
I can see the ODROiD build fail, but i don't think simply editing the vp2430 config could escalate to this: 

`make: *** No rule to make target '3rdparty/dasharo-blobs/hardkernel/odroid_h4/descriptor.bin', needed by 'add_intel_firmware'.  Stop.`

mkopec (approved):

PR #657, opened 2025-04-26
nb/intel/sandybridge/northbridge.c: Disbale non-active PEG devices

miczyg1:
TXT SINIT ACM checks the PEG bridge memory ranges whether they are correctly assigned in the MMIO space. If the bridge is enabled but no device is attached to it, coreboot will not assign any resources (if hotplug is not enabled). When SINIT ACM checks the ranges, it fails on prefetchable range check.

Hide the PEG devices if the bridge is not active. PCIe bridges should generally be hidden if hotplug is not enabled and there is no downstream device.

Upstream-status: TBD

### DIFF
--- a/src/northbridge/intel/sandybridge/northbridge.c
+++ b/src/northbridge/intel/sandybridge/northbridge.c
@@ -267,17 +267,17 @@ static void disable_peg(void)
 	reg = pci_read_config32(dev, DEVEN);
 
 	dev = pcidev_on_root(1, 2);
-	if (!dev || !dev->enabled) {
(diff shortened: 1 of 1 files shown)


miczyg1:
If this works, I will push upstream.

SergiiDmytruk (approved):
s/Disbale/Disable/ for the title, but this does make TXT work.

miczyg1:
> s/Disbale/Disable/ for the title, but this does make TXT work.

Yes, pre-commit caught it when I added the gerrit patch ID to the commit message. It didn't earlier, because I used different PC (which had not pre-commit installed somehow).

PR #582, opened 2024-11-05
mb/msi/ms7d25,7eo6: Fix USB 3.x port mapping

miczyg1:
The USB-C port is Gen2x2 and occupies two USB3.x line pairs. The missing second pair caused all the ports to be shifted by one and port 10 not being enabled as USB3.x capable. As a result one of the JUSB3 ports was working in HighSpeed only.

PR #659, opened 2025-05-05
mb/msi/ms7d25,7eo6: Fix USB 3.x port mapping

miczyg1:
The USB-C port is Gen2x2 and occupies two USB3.x line pairs. The missing second pair caused all the ports to be shifted by one and port 10 not being enabled as USB3.x capable. As a result one of the JUSB3 ports was working in HighSpeed only.

Also revise the USB OC mappings again and put correct OC pins to macros.


miczyg1:
Already reviewed on https://github.com/Dasharo/coreboot/pull/582

PR #662, opened 2025-05-07
MTL TCSS fixes

miczyg1:


miczyg1:
Fixes small misconfiguration, but does not fix the problem with disappearing devices in the dock in Windows

mkopec (approved):
Tested on NovaCustom V540TU with Wavlink UMD05 Pro Rev. E, seems to fix issue https://github.com/Dasharo/dasharo-issues/issues/1316

PR #666, opened 2025-05-12
Nuc fast boot

mkopec:


mkopec:
Fixes https://github.com/Dasharo/dasharo-issues/issues/1321

PR #671, opened 2025-05-15
Novacustom MTL-H S0ix fix

mkopec:
These changes broke SLP_S0 assertion while in S2idle on Linux. Revert to restore S0ix functionality.


filipleple (approved):
plus, this makes the NUC BOX boot reasonably fast. LGTM

PR #686, opened 2025-06-01
soc/intel/fast_spi: fix mapping whole flash for RO

SergiiDmytruk:
This wasn't tested on NovaCustom laptops yet (hence draft status), but MSI has the same issue and an update capsule is able to read start of the flash correctly with this change.

***

This corrects change-id I885252a488bd35fc3afef571e6178642a059f883 ("soc/intel/fast_spi/mmap_boot.c: allow mapping whole flash", 01c93ef).  Looks like hardware/chipset/ME/whatever limits what can be accessed via the memory mapping to IFD BIOS region.

Instead of requesting mapping whatever is in front of IFD BIOS region, add another xlate window for accessing that part of the flash via implementation in src/soc/intel/common/block/fast_spi/fast_spi_flash.c

This is a fix for https://github.com/Dasharo/dasharo-issues/issues/1337.
*ref: ncm-1850*

### DIFF
--- a/src/drivers/efi/Kconfig
+++ b/src/drivers/efi/Kconfig
@@ -42,7 +42,7 @@ config DRIVERS_EFI_MAIN_FW_LSV
 config DRIVERS_EFI_UPDATE_CAPSULES
 	bool "Include EFI update capsules driver"
 	depends on DRIVERS_EFI_VARIABLE_STORE && SMMSTORE_V2 && DRIVERS_EFI_FW_INFO
-	select EXT_BIOS_FILL_UP if SOC_INTEL_COMMON_BLOCK
+	select FAST_SPI_FULL_RO_BOOTMEDIA if SOC_INTEL_COMMON_BLOCK
 	help
 	  Adds a driver that is able to parse CapsuleUpdateData* EFI variables
 	  to discover firmware updates and expose them for tianocore thorough
--- a/src/soc/intel/common/block/fast_spi/Kconfig
+++ b/src/soc/intel/common/block/fast_spi/Kconfig
@@ -52,12 +52,12 @@ config EXT_BIOS_WIN_SIZE
 	  region greater than 16MiB. The actual mapped window might be
 	  smaller depending upon the size of the BIOS region.
 
-config EXT_BIOS_FILL_UP
+config FAST_SPI_FULL_RO_BOOTMEDIA
 	bool
 	default n
 	help
-	  Extend BIOS region to the beginning of the flash for the
-	  purposes of mapping.
+	  Make read-only boot device produced by Fast SPI cover the entire
+	  flash chip.
 
 config FAST_SPI_GENERATE_SSDT
 	bool
--- a/src/soc/intel/common/block/fast_spi/mmap_boot.c
+++ b/src/soc/intel/common/block/fast_spi/mmap_boot.c
@@ -18,15 +18,23 @@ enum window_type {
 	FIXED_DECODE_WINDOW,
 	/* Additional decode window for mapping BIOS region greater than 16MiB */
 	EXT_BIOS_DECODE_WINDOW,
+
+	/* Number of memory-backed regions for reading flash */
+	TOTAL_MMAP_DECODE_WINDOWS,
+
+	/* An optional extra window for part of the flash before IFD BIOS region */
+	PRE_BIOS_DECODE_WINDOW = TOTAL_MMAP_DECODE_WINDOWS,
+
+	/* Maximum number of decode windows */
 	TOTAL_DECODE_WINDOWS,
 };
 
 static struct xlate_region_device real_dev;
-static struct mem_region_device shadow_devs[TOTAL_DECODE_WINDOWS];
+static struct mem_region_device shadow_devs[TOTAL_MMAP_DECODE_WINDOWS];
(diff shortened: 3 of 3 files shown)


philipanda:
@SergiiDmytruk 
I’ve tested manually:
- CUP001.001 Capsule Update With Wrong Keys - results in Security Violation - PASS
- CUP002.001 Capsule Update With Wrong GUID - results in Not Ready - PASS
- CUP150.001 Capsule Update - results in ‘Out of Resources’ - FAIL
Can’t tell for now if the GBE region bug is fixed, because I couldn’t perform the update

The firmware and capsule files I’ve used:
[binaries.zip](https://github.com/user-attachments/files/20685401/binaries.zip)

Capsule is from CI at your commit, firmware was built from the commit after lowering the FW version and LSV in the configs to allow for an update.

Here are some logs, dmidecode, cbmem with flags `-c`, `-1`, `-2`, `-c -B +INFO`
[capsules-logs.zip](https://github.com/user-attachments/files/20685365/capsules-logs.zip)
The update in these logs only ended after a couple of seconds.

The Update does not always end immediately when the 'Out of Resources' status is returned. It sometimes does, but sometimes the progress bar is visible for up to ~3 seconds and progresses to about 1/10 (one square filled?). When the update fails the progress bar skips to the end for a couple miliseconds, hard to notice.

Here are the same logs from an update, that ended immediately:
[capsules_logs_immediate_fail.zip](https://github.com/user-attachments/files/20685610/capsules_logs_immediate_fail.zip)

During my ~7-10 tries of performing a capsule update, I haven't been once able to successfully update the firmware. It's always the 'Out of Resources' status.

philipanda (dismissed):
Tests described in a comment above
https://github.com/Dasharo/coreboot/pull/686#issuecomment-2961406415

SergiiDmytruk:
@philipanda, thanks for the logs.  That's not what I was expecting.  Looks like the firmware (FMP DXE in the capsule) just reboots while trying to verify validity of the capsule, before touching the flash chip.

The most notable difference with MSI is the new DXE for checking AC status.  I tried enabling it for MSI but it worked as before (it doesn't nothing for desktops), so I instead disabled that DXE for NovaCustom in GitHub's CI of this PR.  Maybe give that version of a capsule a try (just `CUP150.001`).

> firmware was built from the commit after lowering the FW version and LSV in the configs to allow for an update.

That shouldn't be necessary.  Updates to the same versions work just fine, it can be non-obvious whether it has worked, but I think we know how a failure looks like here.

mkopec:
Works On My Machine™

Progress bar went to about ~40% and then jumped to 100 and the update ended. Likely because large parts of the firmware were the same? I just reinstalled the same version, so all the update did was essentially wipe ME configuration and MRC cache

This was a bootguard-provisioned binary installed via fwupd cabinet

SergiiDmytruk:
> Works On My Machine™

By that you mean that GBE region was preserved?

> Progress bar went to about ~40% and then jumped to 100 and the update ended. Likely because large parts of the firmware were the same?

Yes, identical blocks are skipped.

> This was a bootguard-provisioned binary installed via fwupd cabinet

So I presume the capsule included AC check DXE?

mkopec:
> By that you mean that GBE region was preserved?

yes, sorry

> So I presume the capsule included AC check DXE?

yes

SergiiDmytruk:
> > By that you mean that GBE region was preserved?
> 
> yes, sorry

Great, thanks.  Then this PR can be reviewed and merged.  "Out of resources" is, apparently, an unrelated issue.

> > So I presume the capsule included AC check DXE?
> 
> yes

Hm, maybe "Out of resources" depends on a model.

mkopec (approved):

mkopec:
@SergiiDmytruk You're missing Upstream-Status

mkopec (changes requested):
@SergiiDmytruk I'm still having some issues sometimes when the MAC is lost. Right now it seems to be working reliably though, so it might be an error with my testing. Will try to narrow it down

mkopec:
@philipanda in your tests, was the USB stick FAT32-formatted or did you use another filesystem? I've seen weird issues like that when using other filesystems despite them being readable

philipanda:
> @philipanda in your tests, was the USB stick FAT32-formatted or did you use another filesystem? I've seen weird issues like that when using other filesystems despite them being readable

It was fat32, check https://github.com/Dasharo/open-source-firmware-validation/blob/novacustom-capsule-tests/scripts/capsules/prepare_capsule_update_tests_drive.sh

mkopec (approved):

SergiiDmytruk:
"Out of resources" is really an issue with reading current ROM image (`CurrentImage = ReadCurrentFirmware ();`), but its cause must be different from reading FFs.  Saw it twice in a row (one fast, another was with a delay) on V540TU but don't have any details because logging (at least to cbmem) seems to just stop somewhere around `Pkcs7Verify()` call.

Checking for reading FFs can be done without logging, but need to flash coreboot with this fix for that.

mkopec:
Managed to reproduce the Out of Resources issue. It reproduces also if I try the same capsule using fwupd

mkopec:
@philipanda can you approve? Your change request is blocking merge. Let's tackle the out of resources issue in another PR

philipanda (commented):
@mkopec you can merge now

PR #697, opened 2025-06-17
config.emulation_qemu_x86_q35_uefi_all_menus: Add CPU Throttling option

philipanda:
QEMU all_menus used to include it just because it was set to true by default.
This was changed in https://github.com/Dasharo/edk2/commit/a5b3b31da521f#diff-fda7be7a2a48a17bb0f3f3b1c3b52b4bc3827f711badc19793e0e840176fbbe7L76-R76 but the option being there was taken for granted by qemu self-tests in OSFV. This PR restores the option to the setup menu to make the OSFV CI self tests on QEMU work again.

I suppose the normal variant of QEMU (not all_menus) shouldn't include the option as I don't think QEMU emulates the CPU overheating :p


macpijan (dismissed):

macpijan:
Thanks for pinpointing the problem!

Please fix check-upstream-status according to https://docs.dasharo.com/dev-proc/source-code-structure/#upstream-status

philipanda:
@macpijan is that updated commit message OK?

macpijan:
Quoting linked documentation:

```
Possible values for reason:

    Dasharo downstream - patch specific to Dasharo distribution
    other explanation - although we should try to avoid it
```

I suggest using the first one, there is typically no need to invent new for each commit.
Also, some checks are still failing here.

mkopec (dismissed):

mkopec (approved):

PR #698, opened 2025-06-20
Odroid btg rebase

filipleple:


mkopec:
why are we dropping the multi-TPM support patches?

filipleple:
> why are we dropping the multi-TPM support patches?

No longer the case, was debugging something earlier

configs/config.hardkernel_odroid_h4_btg
```diff
@@ -0,0 +1,68 @@
...
+CONFIG_POWER_STATE_OFF_AFTER_FAILURE=y
+CONFIG_HAVE_ME_BIN=y
+CONFIG_DASHARO_PREFER_S3_SLEEP=y
+CONFIG_PCIEXP_SUPPORT_RESIZABLE_BARS=y
+CONFIG_DRIVERS_EFI_VARIABLE_STORE=y
+CONFIG_DRIVERS_GENERIC_CBFS_SERIAL=y
+CONFIG_DRIVERS_GENERIC_CBFS_UUID=y
+# CONFIG_DRIVERS_MTK_WIFI is not set
+CONFIG_TPM2=y
+CONFIG_INTEL_TXT_BIOSACM_FILE="3rdparty/dasharo-enterprise-blobs/alderlake_n/ADL_BIOSAC_v1_18_13_20221028_REL_NT_O1.PW_signed_256K.bin"
```
mkopec: on novacustom laptops, the provisioning scripts are responsible for adding the ACMs. Then, there is no new _btg defconfig, users can simply build btg-ready binaries and you eliminate the need for another binary variant. Please take a look at how it's handled there

miczyg1: Exactly. We should not introduce new config and build target. Simply enable Boot Guard in the old ODROID config file.

filipleple: alright with this commit https://github.com/Dasharo/coreboot/pull/698/commits/af9fd1347fb341f961e2c07da0f483a8363eeaf1 and the [corresponding enterprise-blobs pr](https://git.3mdeb.com/dasharo/dasharo-enterprise-blobs/pulls/4) we're using the same approach as the laptops

mkopec (approved):

PR #699, opened 2025-06-20
CBnT: inject measurements made by BootGuard into event log

SergiiDmytruk:
* PCR-0 and (optionally) PCR-7
* TigerLake and newer
* Commit messages and comments in the code provide more details

Was tested for sanity in QEMU which helped to find bugs, but not on hardware (one attempt resulted in a brick although not necessarily related to the new code which doesn't reset the machine or otherwise prevents boot in any way).

*ref: ncm-1811*

### DIFF
--- a/src/arch/x86/include/arch/fit_table.h
+++ b/src/arch/x86/include/arch/fit_table.h
@@ -3,7 +3,10 @@
 #ifndef _FIT_TABLE_H_
 #define _FIT_TABLE_H_
 
+#include <stddef.h>
 #include <stdint.h>
+
+#define FIT_POINTER_ADDRESS      0xffffffc0
 
 #define FIT_ENTRY_HAS_CHECKSUM   0x80
 
@@ -52,5 +55,38 @@ struct fit_entry {
 	uint8_t  type_checksum_valid;
 	uint8_t  checksum;
 } __packed;
+
+static inline uint32_t fit_entry_size(const struct fit_entry *entry)
+{
+	return entry->size_reserved & 0xffffff;
+}
+
+static inline uint8_t fit_entry_type(const struct fit_entry *entry)
+{
+	return entry->type_checksum_valid & 0x7f;
+}
+
+static inline struct fit_entry *fit_table_search(uint8_t entry_type)
+{
+	struct fit_entry *fit = *(struct fit_entry **)(uintptr_t)FIT_POINTER_ADDRESS;
+
+	union {
+		uint8_t bytes[sizeof(uint64_t)];
+		uint64_t u64;
+	} signature = { .bytes = {'_', 'F', 'I', 'T', '_', ' ', ' ', ' '} };
+
+	/* A quick sanity check that we're in fact dealing with FIT. */
+	if (fit->address.u64 != signature.u64)
+		return NULL;
+
+	int entry_count = fit_entry_size(fit) * 16 / sizeof(struct fit_entry);
+
+	for (int i = 0; i < entry_count; i++) {
+		if (fit_entry_type(&fit[i]) == entry_type)
+			return &fit[i];
+	}
+
+	return NULL;
+}
 
 #endif /* _FIT_TABLE_H_ */
--- a/src/arch/x86/include/arch/fit_table.h
+++ b/src/arch/x86/include/arch/fit_table.h
@@ -1,0 +1,56 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+
+#ifndef _FIT_TABLE_H_
+#define _FIT_TABLE_H_
+
+#include <stdint.h>
+
+#define FIT_ENTRY_HAS_CHECKSUM   0x80
+
+/*
+ * From "Firmware Interface Table, BIOS Specification" (document #599500), Rev. 1.6, March 2025
+ */
+#define FIT_ENTRY_TYPE_HEADER    0x00 // FIT Header Entry
+#define FIT_ENTRY_TYPE_UCODE     0x01 // Microcode Update Entry
+#define FIT_ENTRY_TYPE_SACM      0x02 // Startup AC Module Entry
+#define FIT_ENTRY_TYPE_DACM      0x03 // Diagnostic AC Module Entry
+#define FIT_ENTRY_TYPE_PBP       0x04 // Platform Boot Policy Entry
+#define FIT_ENTRY_TYPE_MMF       0x05 // Memory Microcontroller Firmware Entry
+#define FIT_ENTRY_TYPE_FRS       0x06 // FIT Reset State Entry
+#define FIT_ENTRY_TYPE_BIOS_SACM 0x07 // BIOS Startup Module Entry
+#define FIT_ENTRY_TYPE_TPM       0x08 // TPM Policy Record
+#define FIT_ENTRY_TYPE_BIOS      0x09 // BIOS Policy Record
+#define FIT_ENTRY_TYPE_TXT       0x0a // TXT Policy Record
+#define FIT_ENTRY_TYPE_KM        0x0b // Key Manifest Record
+#define FIT_ENTRY_TYPE_BPM       0x0c // Boot Policy Manifest
+#define FIT_ENTRY_TYPE_FSP       0x0d // FSP Boot Manifest
+// 0x0E - 0x0F Intel Reserved
+#define FIT_ENTRY_TYPE_CSE       0x10 // CSE Secure Boot
+// 0x11 - 0x19 Intel Reserved
+#define FIT_ENTRY_TYPE_VAB_PT    0x1a // Vendor Authorized Boot Provisioning Table
+#define FIT_ENTRY_TYPE_VAB_KM    0x1b // Vendor Authorized Boot Key Manifest
+#define FIT_ENTRY_TYPE_VAB_IM    0x1c // Vendor Authorized Boot Image Manifest
+#define FIT_ENTRY_TYPE_VAB_IH    0x1d // Vendor Authorized Boot Image Hash (Deprecated)
+#define FIT_ENTRY_TYPE_VAB_ACPIM 0x1e // VAB Authorization Code Patch Image Manifest
+// 0x1F - 0x2B Intel Reserved
+#define FIT_ENTRY_TYPE_SACM_DBG  0x2c // SACM Debug Record
+#define FIT_ENTRY_TYPE_FPDR      0x2d // Feature Policy Delivery Record.
+				      // Deprecated in CBnT platforms.
+#define FIT_ENTRY_TYPE_SCRTM_ERR 0x2e // Granular SCRTM Error Record
+#define FIT_ENTRY_TYPE_JMP_DBG   0x2f // JMP $ Debug Policy
+// 0x30 - 0x70 Reserved for Platform Manufacturer Use
+// 0x71 - 0x7E Intel Reserved
+#define FIT_ENTRY_TYPE_UNUSED    0x7f // Unused Entry (skip)
+
+struct fit_entry {
+	union {
+		uint8_t bytes[sizeof(uint64_t)];
+		uint64_t u64;
+	} address;
+	uint32_t size_reserved;
+	uint16_t version;
+	uint8_t  type_checksum_valid;
+	uint8_t  checksum;
+} __packed;
+
+#endif /* _FIT_TABLE_H_ */
--- a/src/security/intel/cbnt/measurement.c
+++ b/src/security/intel/cbnt/measurement.c
@@ -3,6 +3,8 @@
 #include <arch/fit_table.h>
 #include <commonlib/iobuf.h>
 #include <console/console.h>
+#include <cpu/intel/msr.h>
+#include <cpu/x86/msr.h>
 #include <device/mmio.h>
 #include <security/intel/cbnt/cbnt.h>
 #include <security/tpm/tspi/crtm.h>
@@ -25,6 +27,13 @@
 #define BPM_IBBS_FLAG_FORCE_TME             (1 << 5)
 #define BPM_IBBS_FLAG_FORCE_IGN             (1 << 6)
 #define BPM_IBBS_FLAG_SRTM_AC               (1 << 7)
+
+/* Bits of `struct km_hash::usage` indicating to which public key the hash corresponds. */
+#define KM_HASH_USAGE_BPM       (1 << 0)
+#define KM_HASH_USAGE_FIT       (1 << 1)
+#define KM_HASH_USAGE_ACM       (1 << 2)
+#define KM_HASH_USAGE_SDEV      (1 << 3)
+#define KM_HASH_USAGE_PFR_CPLD  (1 << 4)
 
 /*
  * Definitions of the structures come from Intel document #575623 CBnT BWG v1.2.5 ("Intel
@@ -336,7 +345,83 @@ static enum cb_err fill_pcr0_km_fields(struct obuf *ob)
 	return err;
 }
 
-static enum cb_err fill_pcr0_bpm_fields(struct obuf *ob)
+/* Finds public key used to sign Key Manifest and hashes it with the algorithm specified in the
+   header. */
+static enum cb_err hash_signature_key(const struct km_header *km, size_t km_len,
+				      struct vb2_hash *hash)
+{
+	/* All safety checks are in skip_signature_key() which is always run for PCR-0, so not
+	   bothering here for optional PCR-7 which is also computed after PCR-0. */
+	struct ibuf sig_ib;
+	ibuf_init(&sig_ib, (uint8_t *)km + km->key_sig_offset, km_len - km->key_sig_offset);
+
+	uint16_t key_alg;
+	(void)ibuf_read_le16(&sig_ib, &key_alg);
+
+	(void)ibuf_oob_drain(&sig_ib, 1); /* key signature version */
+	(void)ibuf_oob_drain(&sig_ib, 2); /* key algorithm */
+	(void)ibuf_oob_drain(&sig_ib, 1); /* key version */
+
+	uint16_t key_size;
+	(void)ibuf_read_le16(&sig_ib, &key_size);
+
+	size_t key_data_size = 0;
+	if (key_alg == TPM_ALG_RSA)
+		key_data_size = 4 + key_size / 8;   /* public exponent and public key */
+	else if (key_alg == TPM_ALG_ECC)
+		key_data_size = (key_size / 8) * 2; /* Qx + Qy */
+
+	if (vb2_hash_calculate(vboot_hwcrypto_allowed(), ibuf_oob_drain(&sig_ib, key_data_size),
+			       key_data_size, tpm2_alg_to_vb2_hash(km->km_pub_key_hash_alg),
+			       hash)) {
+		printk(BIOS_ERR, "CBnT: failed to hash PCR-0 measurement data\n");
+		return CB_ERR;
+	}
+
+	return CB_SUCCESS;
+}
+
+static enum cb_err fill_pcr7_km_fields(struct obuf *ob)
+{
+	size_t km_len;
+	const struct km_header *km = find_in_fit(FIT_ENTRY_TYPE_KM, &km_len);
+
+	struct vb2_hash km_pubkey_hash;
+	if (hash_signature_key(km, km_len, &km_pubkey_hash) != CB_SUCCESS) {
+		printk(BIOS_ERR, "CBnT: failed to hash Key Manifest's signature key\n");
+		return CB_ERR;
+	}
+
+	/* BP.KEY (hash of public part of the key used to sign Key Manifest) */
+	if (obuf_write(ob, km_pubkey_hash.raw, vb2_digest_size(km_pubkey_hash.algo))) {
+		printk(BIOS_ERR,
+		       "CBnT: failed to write hash of Key Manifest's signature key\n");
+		return CB_ERR;
+	}
+
+	/* Find hash that corresponds to BPM. */
+	const struct km_hash *key_hash = km->key_hash;
+	unsigned int i;
+	for (i = 0; i < km->key_count; ++i) {
+		if (key_hash->usage & KM_HASH_USAGE_BPM)
+			break;
+		key_hash = (const struct km_hash *)&key_hash->hash.data[key_hash->hash.size];
+	}
+	if (i == km->key_count) {
+		printk(BIOS_ERR, "CBnT: failed to find hash of BPM pubkey in KM\n");
+		return CB_ERR;
+	}
+
+	/* KM.BPKey.hashBuffer[BPM pubkey digest size] */
+	if (obuf_write(ob, key_hash->hash.data, key_hash->hash.size)) {
+		printk(BIOS_ERR, "CBnT: failed to write ACM signature\n");
+		return CB_ERR;
+	}
+
+	return CB_SUCCESS;
+}
+
+static enum cb_err fill_pcr0_bpm_fields(struct obuf *ob, bool *auth_measure)
 {
 	size_t bpm_len;
 	const struct bpm_header *bpm = find_in_fit(FIT_ENTRY_TYPE_BPM, &bpm_len);
@@ -356,6 +441,11 @@ static enum cb_err fill_pcr0_bpm_fields(struct obuf *ob)
 	}
 
 	const struct bpm_ibbs *ibbs = (const void *)bpm->se_element;
+	*auth_measure = (ibbs->flags & BPM_IBBS_FLAG_AUTH_MEASURE) != 0;
+
+	/* Auth data is measured only for TPM2. */
+	if (*auth_measure && tlcl_get_family() != TPM_2)
+		*auth_measure = false;
 
 	unsigned int i;
 	const struct hash_struct *ibb_hash = ibbs->digest_list.hashes;
@@ -377,6 +467,51 @@ static enum cb_err fill_pcr0_bpm_fields(struct obuf *ob)
 	}
 
 	return CB_SUCCESS;
+}
+
+/*
+ * Pseudo-code of the data to be measured into PCR-7 for TigerLake and newer (older hardware
+ * isn't supported):
+ *
+ *     struct {
+ *         uint64_t ACM_POLICY_STATUS;
+ *         uint16_t ACM.Header.SVN;
+ *         uint8_t  ACM.KeyHash[32];
+ *         uint8_t  BP.KEY[32 or 48];  // Hash of the public key from Key Manifest's signature.
+ *         uint8_t  KM.BPKey.hashBuffer[BPM pubkey digest size];
+ *     } PCR7_DATA
+ *
+ * Applies only for TPM2.0.
+ *
+ * Returns true and initializes *hash on success.
+ */
+static bool make_pcr7_hash(struct obuf *ob, struct vb2_hash *hash)
+{
+	/*
+	 * ACM.KeyHash[32]
+	 *
+	 * BTG BWG says to use TXT.PUBLIC.KEY, but TXT SDG (#315168-017.4) and CBnT BWG say
+	 * that it doesn't contain anything useful on modern platforms utilizing SGX and
+	 * suggests to use MSRs 0x20::0x23.
+	 */
+	(void)obuf_write_le64(ob, rdmsr(MSR_ACM_CPU_KEY_HASH_0).raw);
+	(void)obuf_write_le64(ob, rdmsr(MSR_ACM_CPU_KEY_HASH_1).raw);
+	(void)obuf_write_le64(ob, rdmsr(MSR_ACM_CPU_KEY_HASH_2).raw);
+	(void)obuf_write_le64(ob, rdmsr(MSR_ACM_CPU_KEY_HASH_3).raw);
+
+	if (fill_pcr7_km_fields(ob) != CB_SUCCESS) {
+		printk(BIOS_ERR, "CBnT: failed to fill KM fields of PCR-7 measurement data\n");
+		return false;
+	}
+
+	size_t size;
+	const void *data = obuf_contents(ob, &size);
+	if (vb2_hash_calculate(vboot_hwcrypto_allowed(), data, size, tpm_log_alg(), hash)) {
+		printk(BIOS_ERR, "CBnT: failed to hash PCR-7 measurement data\n");
+		return false;
+	}
+
+	return true;
 }
 
 int intel_cbnt_get_locality(void)
@@ -493,7 +628,8 @@ void intel_cbnt_inject_ibg_measurements(void)
 		return;
 	}
 
-	if (fill_pcr0_bpm_fields(&data_ob) != CB_SUCCESS) {
+	bool auth_measure;
+	if (fill_pcr0_bpm_fields(&data_ob, &auth_measure) != CB_SUCCESS) {
 		printk(BIOS_ERR, "CBnT: failed to fill BPM fields of PCR-0 measurement data\n");
 		return;
 	}
@@ -505,6 +641,20 @@ void intel_cbnt_inject_ibg_measurements(void)
 		return;
 	}
 
+	/* Making and hashing PCR-7 data. */
+	struct vb2_hash pcr7_hash;
+	if (auth_measure) {
+		/* Reuse the first 2 fields of PCR-0 data which are identical in both cases. */
+		obuf_init(&data_ob, data, sizeof(data));
+		(void)obuf_oob_fill(&data_ob, sizeof(uint64_t) + sizeof(uint16_t));
+
+		if (!make_pcr7_hash(&data_ob, &pcr7_hash)) {
+			printk(BIOS_ERR,
+			       "CBnT: failed to build and hash PCR-7 measurement data\n");
+			return;
+		}
+	}
+
 	/*
 	 * BWG suggests to Perform reconstruction at this point, but we are likely to continue
 	 * the boot anyway as Measured Boot failures generally aren't fatal and incorrect hash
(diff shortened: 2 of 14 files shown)


src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	uint16_t hdr_size;
+	uint16_t key_sig_offset;  /* offset to KeySignature field of Policy Manifest Signature
+				     Element (PMSE) */
+	uint8_t bpm_revision;
+	uint8_t bpm_svn;
+	uint8_t acmsvn_auth;
+	uint8_t reserved;
+	uint16_t nem_data_stack;  /* in 4 KiB pages */
+	uint8_t se_element[];
+};
```
miczyg1: coreboot add packed atrribute at the ned of structure definition:

```suggestion
struct bpm_header {
	char structure_id[8];     /* "__ACBP__" */
	uint8_t struct_ver;       /* 0x24 */
	uint8_t hdr_struct_ver;   /* 0x20 */
	uint16_t hdr_size;
	uint16_t key_sig_offset;  /* offset to KeySignature field of Policy Manifest Signature
				     Element (PMSE) */
	uint8_t bpm_revision;
	uint8_t bpm_svn;
	uint8_t acmsvn_auth;
	uint8_t reserved;
	uint16_t nem_data_stack;  /* in 4 KiB pages */
	uint8_t se_element[];
} __packed;
```

Please fix for all other cases

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+#define INTEL_CBNT_CBFS_KM   "key_manifest.bin"
+#define INTEL_CBNT_CBFS_BPM  "boot_policy_manifest.bin"
+
+/* Contains SHA256 hash of OEM's public key. */
+#define OEM_CRED_FILE  "/fpf/intel/OemCred"
+
+/*
+ * Definitions of the structures come from Intel document #575623 CBnT BWG v1.2.5 ("Intel
+ * Converged Boot Guard and Intel Trusted Execution Technology (Intel TXT) BIOS Specification").
+ * It is specific for TigerLake and newer platforms.
```
miczyg1: It is not specific. Structures are backward compatible since introduction of Boot Guard.

Some fields were deprecated, but the yare simply marked as reserved.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	uint8_t struct_ver;    /* 0x21 */
+	uint8_t reserved1;
+	uint16_t element_size;
+	uint8_t reserved2;
+	uint8_t set_number;    /* 0 */
+	uint8_t reserved3;
+	uint8_t pbet_value;
+	uint32_t flags;        /* see BPM_IBBS_FLAG_* constants for values */
+	uint64_t bar0;
+	uint64_t bar1;
```
miczyg1: These are VTD/IOMMU BARs.
```suggestion
	uint64_t iommu_bar0;
	uint64_t iommu_bar1;
```

src/soc/intel/common/block/cse/cse.c
```diff
@@ -933,6 +935,58 @@ void print_me_fw_version(void *unused)
...
+	strncpy(msg.file_name, path, sizeof(msg.file_name));
+	msg.hdr.group_id = MKHI_GROUP_ID_MCA;
+	msg.hdr.command = MKHI_READ_FILE;
+	msg.flags = flags;
+	msg.data_size = *size;
+	msg.offset = offset;
+
+	reply_size = sizeof(rmsg);
+
+	if (heci_send_receive(&msg, sizeof(msg), &rmsg, &reply_size, HECI_MKHI_ADDR)) {
```
miczyg1: This function should check the state of ME, not sure if reading a file is allowed in every state of ME.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+{
+	return cbfs_unmeasured_area_map("COREBOOT", name, size_out);
+}
+
+static enum cb_err fill_pcr0_acm_fields(struct obuf *ob)
+{
+	/* Not running any checks on the ACM like TXT code does, it must be correct if we got
+	   here. */
+	size_t acm_len;
+	struct acm_header_v0 *acm = map_cbfs_file(CONFIG_INTEL_TXT_CBFS_BIOS_ACM, &acm_len);
```
miczyg1: I think we should not introduce yet a new API for CBFS access.

Instead, I would simply reference ACM and manifests from FIT directly: https://www.intel.com/content/dam/develop/external/us/en/documents/firmware-interface-table-bios-specification-r1p2p1.pdf

Because we inject manifests as a last step, they cannot be verified with CBFS_VERIFICATION to avoid chicken and egg problem with metadata hash.

There is another reason to prefer the FIT method. If we introduce top swap, then we will have to change the region where to look for ACMs and manifests probably. With FIT method, that would not be necessary as FIT pointer woudl always point to the ACMs and manifests used in current boot.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+		}
+	}
+	if (key_alg == TPM_ALG_ECC) {
+		if (sig_scheme != TPM_ALG_ECDSA && sig_scheme != TPM_ALG_SM2) {
+			printk(BIOS_ERR, "CBnT: unexpected public signature scheme: %#04x\n",
+			       sig_scheme);
+			return CB_ERR;
+		}
+	}
+
```
miczyg1: If we try to be so thorough with checking, then I would suggest using `else if` here and also an `else` with `unrecognized key algorithm` error and CB_ERR return.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	 * ACM.KeyHash[32]
+	 *
+	 * BWG says to use TXT.PUBLIC.KEY, but TXT SDG (#315168-017.4) says that it
+	 * doesn't contain anything useful on modern platforms and suggests SGX-related
+	 * MSRs (0x20::0x23, to be specific, but SDM vol. 4 has these instead).
+	 */
+	(void)obuf_write_le64(ob, rdmsr(MSR_IA32_SGXLEPUBKEYHASH0).raw);
+	(void)obuf_write_le64(ob, rdmsr(MSR_IA32_SGXLEPUBKEYHASH1).raw);
+	(void)obuf_write_le64(ob, rdmsr(MSR_IA32_SGXLEPUBKEYHASH2).raw);
+	(void)obuf_write_le64(ob, rdmsr(MSR_IA32_SGXLEPUBKEYHASH3).raw);
```
miczyg1: Hmm. this all looks weird. Also these SGX registers do not exists on all platforms, especially that SGX has been long gone on client machines (last time it is present on CometLake, should not be available on Tiger Lake), onyl exists on servers IIRC.

I did a quick test on my ADL employee laptop with BTG enabled:

```
[miczyg@3M08 coreboot]$ sudo mmio_dump 0xfed30400 0x20
0x00000000fed30400: 0x9c8f9a87 0x1d3d9ebf 0xd79adc12 0xe634de6d
0x00000000fed30410: 0x643640aa 0x34db39c7 0x0b8f857b 0x3aae33e0
[miczyg@3M08 coreboot]$ sudo rdmsr 0x20
350eec1f9866521
[miczyg@3M08 coreboot]$ sudo rdmsr 0x21
de1f549fc0756be
[miczyg@3M08 coreboot]$ sudo rdmsr 0x22
8e2a605e2b89849d
[miczyg@3M08 coreboot]$ sudo rdmsr 0x23
79bd6994c6a790c1
[miczyg@3M08 coreboot]$ sudo rdmsr 0x8c
rdmsr: CPU 0 cannot read MSR 0x0000008c
[miczyg@3M08 coreboot]$ sudo rdmsr 0x8d
rdmsr: CPU 0 cannot read MSR 0x0000008d
[miczyg@3M08 coreboot]$ sudo rdmsr 0x8e
rdmsr: CPU 0 cannot read MSR 0x0000008e
[miczyg@3M08 coreboot]$ sudo rdmsr 0x8f
rdmsr: CPU 0 cannot read MSR 0x0000008f
```

So the MSRs 0x20:0x23 and TXT.PUBLIC.KEY seem to be different. Now to confirm which method is correct I sha256sum'ed the RSA key in ACM header:

```
dd if=ADL_BIOSAC_v1_18_13_20221028_REL_NT_O1.PW_signed_256K.bin skip=128 count=1 bs=384 iflag=skip_bytes |sha256sum
1+0 records in
1+0 records out
384 bytes copied, 5.0509e-05 s, 7.6 MB/s
216586f9c1ee5003be5607fc49f5e10d9d84892b5e602a8ec190a7c69469bd79  -
```

The hash obtained from the ACM header matches what is in the MSR 0x20:0x24 (of course in different endianess).

tl;dr

1. Can't use SGX MSRs on TGL+. Also these MSRs probably contain the enclave key, not ACM key.
2. Can't use TXT.PUBLIC.KEY as BWG suggests.
3. Should use MSRs 0x20:0x23 as BWG suggests.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+
+	/* BP.KEY[32] (rather its SHA256 hash) */
```
miczyg1: Hash algorithm of this file is microachitecture dependent. TGL and onward already provision ME with SHA384 of the key, so the hash read from ME will be SHA384.

```suggestion
	/* BP.KEY[32] (rather its hash) */
```

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	uint8_t bp_key[32];
+	size_t sz = sizeof(bp_key);
+	if (!read_cse_file(OEM_CRED_FILE, bp_key, &sz, 0, MKHI_READ_FILE_FLAG_DEFAULT)) {
```
miczyg1: If ME is not available this command may fail. As a fallback we may read out that key from Key Manifest and hash it with proper algorithm (TGL and onwards SHA384, prior to TGL SHA256). It is better to avoid involving ME if not necessary. We have poor control over it, so it would be best to utilize what we built ourselves (i.e the manfiests).

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	if (biosacm_sts.status.scrtm_status == 0)
+		return;
+
+	/* Do nothing if BootGuard doesn't measure anything. */
+	if (!(biosacm_sts.status.bp & CBNT_BP_TYPE_M))
+		return;
+
+	/*
+	 * At this point the code should check whether PCRs 0-7 were capped with a digest of
+	 * value "1" (ASCII? binary? hex?) and if so, add appropriate entries to the event log.
```
miczyg1: The ACM also caps a bank if a bank is enabled bu the appropriate digest of IBB is not in the BPM. E.g. SHA21 enabled bu no SHA1 digest of IBB in BPM, SHA1 bank is capped.

Also the code may simply check TPMFailureAction in BPM IBB Segment Element flags to see what action ACM takes on error. And whether a TPM error occurred, check ACM_POLICY_STATUS bit 15 or the SACM_INFO MSR bit 3.

BWG says that BIOS should log EV_SEPARATOR in the case of capping. It can be found here: https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf

```
Per Section 3.3.2 Error Conditions, used to
indicate an error occurred recording the
CRTM, POST BIOS, or Embedded Option
ROMs. For this situation, the digest field
MUST contain the tagged digest of the value
00000001h. The event field is arbitrary to
allow platform manufacturers to include an
indication of what error occurred or other
useful troubleshooting information.
```

So it is a 32bit hex value of "1" that is hashed

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	 * This can only happen if Bit3 of IbbFlags is set ("On error: Leave TPM Hierarchies
+	 * enabled. Cap all PCRs"), otherwise TPM is just disabled.  Either way Measured Boot
+	 * won't work, so skipping it unless reasons for doing it are discovered.
+	 */
+
+	/*
+	 * Making and hashing PCR-7 data.
+	 *
+	 * Pseudo-code of the data to be measured into PCR-0 for TigerLake and newer (older
+	 * hardware isn't supported):
```
miczyg1:
```suggestion
	 * hardware isn't supported yet):
```

But the process is pretty much the same (just slightly different way of obtaining the data).

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+	 * Allow for 4 8192-bit digests in PCR-0 measurements which is an unlikely situation,
+	 * so this buffer should be enough to not run out of space (data measured into PCR-7 is
+	 * smaller).
+	 */
+	char data[sizeof(uint64_t) + sizeof(uint16_t) + 4 * KiB];
+	struct obuf data_ob;
+	obuf_init(&data_ob, data, sizeof(data));
+
+	/* ACM_POLICY_STATUS (won't run out of space on this one) */
+	(void)obuf_write_le64(&data_ob, biosacm_sts.raw);
```
miczyg1, line 619: Why the `(void)` at the beginning?

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,543 @@
...
+			return;
+		}
+	}
+
+	/*
+	 * BWG suggests to Perform reconstruction at this point, but we are likely to continue
+	 * the boot anyway as Measured Boot failures generally aren't fatal and incorrect hash
+	 * will be caught later during the boot process.  So not bothering.
+	 */
+
```
miczyg1, line 663: I think we are also missing the `10.4.5.3 Startup Locality Event` https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf

if TPM is started from locality 3 instead of 0, then such information must be recorded in the event log.

```
The Startup Locality event should be placed in the log before any event which extends PCR[0]. This allows
software which needs to parse the TCG Event Log to initialize its internal PCR[0] state correctly. See Section
3.3.4.1 PCR[0] – SRTM, POST BIOS, and Embedded Drivers.
```

The startup locality is also controlled with a BPM flag and the locality used to perform startup is available in ACM_POLICY_STATUS bit 36

src/security/tpm/tspi/log-tpm2.c
```diff
@@ -22,23 +22,6 @@
...
-		return TPM2_ALG_SHA384;
-	case VB2_HASH_SHA512:
-		return TPM2_ALG_SHA512;
-
-	default:
-		return 0xFF;
-	}
-}
-
 void *tpm2_log_cbmem_init(void)
```
miczyg1, line 34: One more thing to that function. The log size should be 64K at minimum per the TCG spec:

```
4. If the ACPI TPM2 table contains the address and size of the Platform Firmware TCG log:
a. The Log Area Minimum Length for the TCG event log MUST be at least 64KB.
```

Here we just allocate 50 entries + size of the log table. Same happens of TPM1.2. Because of that the platform ends up having just a few hundreds of bytes of the log in ACPI TPM table. Not sure if measured boot v2 PRs change that on EDK2 side somehow. Still the EDK2 is not the only payload we use.

miczyg1 (changes requested):

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Moved: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R60

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Removed that line: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7L49

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Renamed: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R86

src/security/intel/cbnt/measurement.c
SergiiDmytruk: The check for unexpected value is in `skip_bpm_signature_key()` called above, so `key_alg` can't be anything but these two values.  But updated to use `else if` for clarity: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R259

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Thanks, updated: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R482

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Updated: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R534

src/security/intel/cbnt/measurement.c
SergiiDmytruk: To state that the result of the invocation is being discarded intentionally.

src/security/tpm/tspi/log-tpm2.c
SergiiDmytruk: coreboot doesn't actually store measured data in the events, so it's enough for coreboot.  Measured boot v2 stops coreboot from publishing its event log in ACPI so EDK can publish its version which is much larger than what coreboot uses.

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Switched to parsing FIT: https://github.com/Dasharo/coreboot/pull/699/commits/49a4b71ffd36822ff25b22469fcf3d7cd9cbf21d and https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R129.

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Updated: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R376

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Hashing part of Key Manifest now: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R330

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Updated  the comment, didn't introduce querying active banks for handling capping: https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R534

src/soc/intel/common/block/cse/cse.c
SergiiDmytruk: It's no longer there (dropped that commit): https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-83770e2316e240c6b33f5ea53347c30dc46faeadcf27d405f50b38562f61b233L974

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Iimplemented, should work if whatever processes it allows event's data to be larger than expected: https://github.com/Dasharo/coreboot/pull/699/commits/24f501d37a12c56db9a38321402e32ca43e8a618 and https://github.com/Dasharo/coreboot/compare/ffa516f0285a87e049c9e7ebefc0b23e3d941a77..4b501619643dc15f6eb104ad64919f0fb0b5e347#diff-2191f194b15fa678d26fbe3aa4c3df04b08093f744b0964aca8e2270abcb7910.

src/security/tpm/tspi/log-tpm2.c
miczyg1: > Measured boot v2 stops coreboot from publishing its event log in ACPI so EDK can publish its version which is much larger than what coreboot uses.

yeah, I know, but when EDK2 is not used, the log ends up being tiny.

src/security/intel/cbnt/measurement.c
miczyg1: >  didn't introduce querying active banks for handling capping

Leave a TODO then?

src/arch/x86/include/arch/fit_table.h
```diff
@@ -0,0 +1,82 @@
...
+static inline uint8_t fit_entry_type(const struct fit_entry *entry)
+{
+	return entry->type_checksum_valid & 0x7f;
+}
+
+static inline struct fit_entry *fit_table_search(uint8_t entry_type)
+{
+	struct fit_entry *fit = *(struct fit_entry **)(uintptr_t)FIT_POINTER_ADDRESS;
+	int entry_count = fit_entry_size(fit) * 16 / sizeof(struct fit_entry);
+
```
miczyg1, line 83: Shouldn't we check if the first FIT entry has the proper signature?

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,621 @@
...
+ * Returns true and initializes *hash on success.
+ */
+static bool make_pcr7_hash(struct obuf *ob, struct vb2_hash *hash)
+{
+	/*
+	 * ACM.KeyHash[32]
+	 *
+	 * BWG says to use TXT.PUBLIC.KEY, but TXT SDG (#315168-017.4) says that it
+	 * doesn't contain anything useful on modern platforms and suggests SGX-related
+	 * MSRs (0x20::0x23, to be specific, but SDM vol. 4 has these instead).
```
miczyg1: Where is that suggestion in SDM? I think this comment should be removed/changed.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,621 @@
...
+	void *addr = (void *)(uintptr_t)entry->address.u64;
+
+	switch (type) {
+	case FIT_ENTRY_TYPE_SACM:
+		if (fit_entry_type(&entry[1]) == FIT_ENTRY_TYPE_SACM) {
+			/* A BIOS image can include more than Startup ACM, but this function
+			   always returns the first one which might be wrong. */
+			printk(BIOS_WARNING, "CBnT: picking SACM from FIT at random!\n");
+		}
+		*size = ((struct acm_header_v0 *)addr)->size * 4;
```
miczyg1: The `acm_header_v0` might need an update. There are newer versions of acm_header. I believe ADL or MTL uses v3 already which has slightly different structure (RSAPubKey/RSASig is bigger, depends on KeySize field and RSAPubExp is omitted, Scratch and User Area also get shifted). SO maybe a union of acm header versions would be needed.

src/arch/x86/include/arch/fit_table.h
SergiiDmytruk: Added: https://github.com/Dasharo/coreboot/compare/4f0d0aa19a293ff4329d846c6f39af07d848a04c..3eb202f68e21e208692f41e80043eddf08d69588#diff-b3cae2b23269e1bd1b562bf387e9bc0017526cd57b2979919b9e36995a9addbc

src/security/tpm/tspi/log-tpm2.c
SergiiDmytruk: Right, but then nothing except coreboot likely writes to the log.

src/security/intel/cbnt/measurement.c
SergiiDmytruk: `B.1.21 TXT.PUBLIC.KEY – AC Module Public Key Hash`:

> Note: Modern platforms utilizing SGX technology implement the Public Key Hash register as
part of the CPU core, not as part of PCH and thus TXT.PUBLIC.KEY register has been
deprecated with no underlying HW functionality. Any value it may contain does not have
meaning and must be disregarded.
Actual Public Key Hash is typically available in CPU MSRs 0x20∷0x23.
Platform manufacturers shall consult respective external data sheets to verify the above
numeric values as it may change.

Also mentioned when `EVTYPE_SINIT_PUBKEY_ HASH` is described:
> PUBKEY_HASH is digest of ACM Signer key. It can be retrieved from MSRs 0x20::0x23

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Implemented capping on error and left a TODO about active PCRs: https://github.com/Dasharo/coreboot/compare/4f0d0aa19a293ff4329d846c6f39af07d848a04c..3eb202f68e21e208692f41e80043eddf08d69588#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R520-R580

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Key data is at different offsets indeed, thanks.

v3 should cover our needs, it's also the first version with CBnT support.  TGL and ADL use v3, MTL has v4.  The signature part is compatible between v3.0 and v5.4 (v5 seems to have no change for the header compared to v4, just version number change).

code changes: https://github.com/Dasharo/coreboot/compare/4f0d0aa19a293ff4329d846c6f39af07d848a04c..3eb202f68e21e208692f41e80043eddf08d69588#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R145-R187
declaration: https://github.com/Dasharo/coreboot/compare/4f0d0aa19a293ff4329d846c6f39af07d848a04c..3eb202f68e21e208692f41e80043eddf08d69588#diff-1013ce554074c41208e8c9b5239fbe83423c40bf24d9a07ff5b98440bba4a74c

SergiiDmytruk (commented):
Other changes apart from rebasing and review:
 * removed Q35 and CBFS leftovers
 * made more things `const` which wasn't possible due to CBFS
 * added error checks after FIT lookup since `find_in_fit()` is returning `NULL` on error

src/security/tpm/tspi/log-tpm2.c
miczyg1: @SergiiDmytruk SeaBIOS does IIRC.

src/security/intel/cbnt/measurement.c
miczyg1: That's a note from MLE SDG. SDM is a different document. But I already see where it came from.

1. MLE SDG and CBNT spec are right when it comes to MSRs 0x20:0x23, which I already proved.
2. We have to remove any reference of SDM here. It seems you have mistaken SGXLEPUBKEYHASH with ACM PUBKEYAHSH MSRs, because both have PUBKEYHASH in its names. But the ACM PUBKEY MSRS are nto SGX-related at all. Just the information about ACM pubkey hash has been moved from PCH MMIO registers to CPU MSRs since the introduction of SGX technology. However, SGX has been removed after Comet Lake (10th Gen) from client devices, but the MSRs 0x20:0x23 are still there, because they have different purpose and are not SGX related.

Regardless of the reason, SDM does not have any information about Boot Guard event logging, nor about MSRS 0x20:0x23. Thus the comment should be rephrased and not mention SDM at all:

```
	 * BTG BWG says to use TXT.PUBLIC.KEY, but TXT SDG (#315168-017.4) and CBnT BWG
	 * says that it doesn't contain anything useful on modern platforms utilizing SGX and
	 * suggests to use MSRs 0x20::0x23.
```

One may also calculate the hash of the key from ACM header itself and compare against TXT.PUBLIC.KEY or the MSRs to be sure. But I think the Skylake+ (6th gen) will have the key in MSRs already.

src/security/intel/cbnt/measurement.c
miczyg1: Is it good that we assume V3 always? Maybe a union would be better + a check for ACM header version.

src/security/tpm/tspi/log-tpm2.c
SergiiDmytruk: It does (discarding anything that coreboot wrote there), although it respects the length reported in ACPI.  Added the change anyway: https://github.com/Dasharo/coreboot/pull/699/commits/84fb089a040e938ced09b2d74d1b5dd9d17e3348.

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Added a version check in https://github.com/Dasharo/coreboot/compare/5820a99a5ee7959893db963edf3b597bc5e63554..84fb089a040e938ced09b2d74d1b5dd9d17e3348#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7.

I wanted to add a union last time, but it wouldn't do anything useful in this case (v4/v5 adds and modifies some fields at the end), so I didn't even bother committing the header for v4.

src/security/intel/cbnt/measurement.c
SergiiDmytruk: > 2\. It seems you have mistaken SGXLEPUBKEYHASH with ACM PUBKEYAHSH MSRs, because both have PUBKEYHASH in its names.

That wasn't a mistake.  I picked it because I didn't find anything that fits the description nor description of MSRs 0x20:0x23 in SDM, and since SGX was mentioned in both places decided to go with it which is what the comment was explaining.  I should have updated it when switching to 0x20:0x23, did it now: https://github.com/Dasharo/coreboot/compare/5820a99a5ee7959893db963edf3b597bc5e63554..84fb089a040e938ced09b2d74d1b5dd9d17e3348#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7R494

src/security/tpm/tspi/log-tpm2.c
miczyg1: Thank you

src/security/intel/cbnt/measurement.c
miczyg1: > I picked it because I didn't find anything that fits the description nor description of MSRs 0x20:0x23 in SDM

Still don't understand why using SDM here, but never mind.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,675 @@
...
+		printk(BIOS_ERR, "CBnT: failed to find FIT entry: %#02x\n", type);
+		return NULL;
+	}
+
+	void *addr = (void *)(uintptr_t)entry->address.u64;
+
+	switch (type) {
+		/* All known versions of the header (up to v5.4) have compatible version and
+		   size fields. */
+		const struct acm_header_v0 *acm;
```
miczyg1: Variable declarations should be on top of the function content block.

src/security/intel/cbnt/measurement.c
```diff
@@ -0,0 +1,675 @@
...
+
+	case FIT_ENTRY_TYPE_SACM:
+		if (fit_entry_type(&entry[1]) == FIT_ENTRY_TYPE_SACM) {
+			/* A BIOS image can include more than Startup ACM, but this function
+			   always returns the first one which might be wrong. */
+			printk(BIOS_WARNING, "CBnT: picking SACM from FIT at random!\n");
+		}
+
+		acm = (const struct acm_header_v0 *)addr;
+		if (acm->header_version[1] > 5 || acm->header_version[0] > 4) {
```
miczyg1, line 150: The important comment wandered to the top of switch statement, but was probably meant to be here:

```
		/* All known versions of the header (up to v5.4) have compatible version and
		   size fields. */
```

src/security/intel/cbnt/measurement.c
SergiiDmytruk: > Still don't understand why using SDM here

Because Volume 4 of Intel 64 and IA-32 Architectures Software Developer’s Manual describes MSRs.

miczyg1 (dismissed):
Minor code style problems left

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Moved the comment here: https://github.com/Dasharo/coreboot/compare/84fb089a040e938ced09b2d74d1b5dd9d17e3348..4a20083b3990dba0883f92b1f90408bbbb276a61#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7

src/security/intel/cbnt/measurement.c
SergiiDmytruk: Moved the declaration before the `switch`: https://github.com/Dasharo/coreboot/compare/84fb089a040e938ced09b2d74d1b5dd9d17e3348..4a20083b3990dba0883f92b1f90408bbbb276a61#diff-1725d4cd7234cc446584ac6ebad791ed8c8e640970d4adb2e99e5b5b233030f7

Top of the function would be farther away from the use and inconsistent with other declarations in this and other functions.

mkopec (approved):

macpijan:
@miczyg1 I would appreciate dropping the `request change` status if your comments  are addressed.

EDIT:

Ok @mkopec could also do that for these style fixes.

PR #700, opened 2025-06-22
Fix RW window for accessing full flash

SergiiDmytruk:
* Make several attempts to enable full flash access in `enable_capsule_smi()`.  This prevents a situation when capsule update fails right away due to `EFI_UNSUPPORTED` on first attempt to communicate with SMMSTORE.
* Bump EDK to get a related fix there.
* Fix RW window in fast SPI flash code.

This is part of fixing https://github.com/Dasharo/dasharo-issues/issues/1338.

EDK part: https://github.com/Dasharo/edk2/pull/255

*ref: ncm-1850*

### DIFF
--- a/payloads/external/edk2/Kconfig.dasharo
+++ b/payloads/external/edk2/Kconfig.dasharo
@@ -4,7 +4,7 @@ config EDK2_REPOSITORY
 	default "https://github.com/Dasharo/edk2"
 
 config EDK2_TAG_OR_REV
-	default "e4dea91cef0af9dbb05e7a407bb18ca18080cbaf"
+	default "22333c9e249641293ce4415e726354f396a0d8d3"
 
 config EDK2_SYSTEM76_EC_LOGGING
 	bool "Enable edk2 logging to System76 EC"
--- a/src/drivers/efi/capsules.c
+++ b/src/drivers/efi/capsules.c
@@ -7,6 +7,7 @@
 #include <cbmem.h>
 #include <console/console.h>
 #include <cpu/x86/pae.h>
+#include <delay.h>
 #include <drivers/efi/efivars.h>
 #include <drivers/efi/capsules.h>
 #include <memrange.h>
@@ -792,8 +793,20 @@ static void enable_capsule_smi(void *unused)
 {
 	uint32_t ret;
 
-	ret = call_smm(APM_CNT_SMMSTORE, SMMSTORE_CMD_USE_FULL_FLASH,
-		       (void*)(uintptr_t)uefi_capsule_count);
+	/* SMI can occasionally be ignored, so retry several times on failure. */
+	uint8_t retries_left = 10;
+	while (1) {
+		ret = call_smm(APM_CNT_SMMSTORE, SMMSTORE_CMD_USE_FULL_FLASH,
+			       (void *)(uintptr_t)uefi_capsule_count);
+		if (ret == SMMSTORE_RET_SUCCESS)
+			break;
+
+		if (retries_left-- == 0)
+			break;
+
+		/* Add a tiny delay before making another try. */
+		udelay(100);
+	}
 
 	printk(BIOS_INFO, "%sabled capsule update SMI handler\n",
 	       ret == SMMSTORE_RET_SUCCESS ? "En" : "Dis");
--- a/src/soc/intel/common/block/fast_spi/mmap_boot.c
+++ b/src/soc/intel/common/block/fast_spi/mmap_boot.c
@@ -22,28 +22,55 @@ enum window_type {
 	/* Number of memory-backed regions for reading flash */
 	TOTAL_MMAP_DECODE_WINDOWS,
 
-	/* An optional extra window for part of the flash before IFD BIOS region */
-	PRE_BIOS_DECODE_WINDOW = TOTAL_MMAP_DECODE_WINDOWS,
-
-	/* Maximum number of decode windows */
+	/*
+	 * Maximum number of decode windows.
+	 *
+	 * There is an optional RW window to cover part of the flash preceding IFD BIOS region,
+	 * so `TOTAL_DECODE_WINDOWS` is `TOTAL_MMAP_DECODE_WINDOWS + 1`.
+	 */
 	TOTAL_DECODE_WINDOWS,
 };
 
+static struct mem_region_device mmap_shadow_devs[TOTAL_MMAP_DECODE_WINDOWS];
+static struct xlate_window mmap_windows[TOTAL_MMAP_DECODE_WINDOWS];
+
+static struct xlate_window rw_window;
+
 static struct xlate_region_device real_dev;
-static struct mem_region_device shadow_devs[TOTAL_MMAP_DECODE_WINDOWS];
 static struct xlate_window real_dev_windows[TOTAL_DECODE_WINDOWS];
+static size_t real_dev_win_count;
+
+static void append_real_dev_window(const struct xlate_window *w)
+{
+	if (real_dev_win_count < ARRAY_SIZE(real_dev_windows))
+		real_dev_windows[real_dev_win_count++] = *w;
+	else
+		printk(BIOS_CRIT, "Fast SPI BUG: no room to add a window, ignoring it\n");
+}
 
 static void initialize_mmap_window(enum window_type type, uintptr_t host_base,
 				   uintptr_t flash_base, size_t size)
 {
-	mem_region_device_ro_init(&shadow_devs[type], (void *)host_base, size);
-	xlate_window_init(&real_dev_windows[type], &shadow_devs[type].rdev,
-			  flash_base, size);
+	mem_region_device_ro_init(&mmap_shadow_devs[type], (void *)host_base, size);
+	xlate_window_init(&mmap_windows[type], &mmap_shadow_devs[type].rdev, flash_base, size);
 	printk(BIOS_INFO, "%s: ",
 		(type == FIXED_DECODE_WINDOW) ?
 		 "Fixed Decode Window" : "Extended Decode Window");
 	printk(BIOS_INFO, "SPI flash base=0x%lx, Host base=0x%lx, Size=0x%zx\n",
 	       flash_base, host_base, size);
+
+	append_real_dev_window(&mmap_windows[type]);
+}
+
+static void initialize_rw_window(uintptr_t flash_base, size_t size)
+{
+	const struct region_device *rw_device = boot_device_rw();
+
+	xlate_window_init(&rw_window, rw_device, flash_base, size);
+	printk(BIOS_INFO, "Fast SPI RW Decode Window: SPI flash base=0x%lx, Size=0x%zx\n",
+	       flash_base, size);
+
+	append_real_dev_window(&rw_window);
 }
 
 /*
@@ -100,8 +127,6 @@ static void bios_mmap_init(void)
 	uintptr_t ext_win_host_base, ext_win_flash_base;
 	size_t fixed_win_size, ext_win_size;
 
-	size_t win_count = 0;
-
 	if (init_done)
 		return;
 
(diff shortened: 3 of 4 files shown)


src/drivers/efi/capsules.c
```diff
@@ -792,8 +793,20 @@ static void enable_capsule_smi(void *unused)
 {
 	uint32_t ret;
 
-	ret = call_smm(APM_CNT_SMMSTORE, SMMSTORE_CMD_USE_FULL_FLASH,
-		       (void*)(uintptr_t)uefi_capsule_count);
+	/* SMM can occasionally ignore SMI calls, so retry several times on failure. */
+	uint8_t retries_left = 10;
+	while (1) {
```
philipanda, line 798: Purely visual suggestion, but I find it weird to use an infinite loop when the number of iterations is clearly capped by `retries_left`.

Why not do
```c
while (retries_left != 0)
```
and replace the lines
 ```c
if (retries_left-- == 0)
  break;
```
with
```c
retries_left--;
```

philipanda (dismissed):
Only an insignificant subjective style suggestion

src/drivers/efi/capsules.c
SergiiDmytruk: Then it will be number of tries and the last attempt will be followed by a delay.  Although the condition can be checked as you suggest while omitting the delay, can do that.

SergiiDmytruk (commented):
This PR doesn't fix the issue as intended, so marking as a draft for now.

src/drivers/efi/capsules.c
philipanda: Right, I've missed that. In that case the first call of `call_smm` could be moved in front of the loop, and then the loop could first check the return value, then wait the delay, and retry at the end of an iteration. Anyway, I don't think it's important and only subjective

krystian-hebel:
Not much useful info could be obtained from capsule update process, it just hanged somewhere inside SMI handler, after printing input parameters, already on the first read. They looked mostly OK, but reported `param` address _could_ be in different than it should be. The address was too convincing to be disregarded as EC debugger garbage. I don't have the logs from that though, it was obtained on @mkopec's laptop. We could try adding more debug output to `rdev_readat()` or even raw SPI driver calls to try to pinpoint where exactly it hangs, but this is time consuming.

I do have log from OS booted normally (dirty because I also did https://review.coreboot.org/c/coreboot/+/86038, didn't help):
[nvc.log](https://github.com/user-attachments/files/21127275/nvc.log)

Interesting points:
- There are 60 `SMI_STS: APM` lines, 54 of them are preceded by double (!?) empty `SMI_STS: TCO` entries.
- In another entry, just before those 54, between two TCOs there are logs from SMMSTORE search and nothing else, so it should be counted as 55th set of TCO -> TCO -> APM entries.
- According to https://github.com/Dasharo/coreboot/blob/e2e2acc0a045f283e24de04743a85e3d8fbf1c23/src/soc/intel/common/block/smm/smihandler.c#L443, one TCO could exist, no idea where the second one comes from.
- Something strange happens along the way, SMMSTORE is reported to be cleared 5 times times during boot, in some cases a write happens immediately after. It this is normal, I'm surprised we haven't wore out flash yet. Example:

<details>

```
[NOTE ]  coreboot-dell_optiplex_7010_9010_v0.1.2-rc1-133-gaa3639ed8cec-dirty-v1.0.0-rc3 Sun Jul 06 13:45:35 UTC 2025 x86_32 smm starting (log level: 0)...
[DEBUG]  SMI_STS: TCO 


[NOTE ]  coreboot-dell_optiplex_7010_9010_v0.1.2-rc1-133-gaa3639ed8cec-dirty-v1.0.0-rc3 Sun Jul 06 13:45:35 UTC 2025 x86_32 smm starting (log level: 0)...
[DEBUG]  SMI_STS: TCO 


[NOTE ]  coreboot-dell_optiplex_7010_9010_v0.1.2-rc1-133-gaa3639ed8cec-dirty-v1.0.0-rc3 Sun Jul 06 13:45:35 UTC 2025 x86_32 smm starting (log level: 0)...
[DEBUG]  SMI_STS: APM 
[DEBUG]  Raw clear SMM store, param = 0x493d6000
[DEBUG]  FMAP: area SMMSTORE found @ 1020000 (262144 bytes)


[NOTE ]  coreboot-dell_optiplex_7010_9010_v0.1.2-rc1-133-gaa3639ed8cec-dirty-v1.0.0-rc3 Sun Jul 06 13:45:35 UTC 2025 x86_32 smm starting (log level: 0)...
[DEBUG]  SMI_STS: TCO 
[WARN ]  EFIVARS: No Firmware Volume header present
[WARN ]  EFIVARS: Failed to validate firmware header
[WARN ]  EFIVARS: No Firmware Volume header present
[WARN ]  EFIVARS: Failed to validate firmware header


[NOTE ]  coreboot-dell_optiplex_7010_9010_v0.1.2-rc1-133-gaa3639ed8cec-dirty-v1.0.0-rc3 Sun Jul 06 13:45:35 UTC 2025 x86_32 smm starting (log level: 0)...
[DEBUG]  SMI_STS: TCO 
[WARN ]  EFIVARS: No Firmware Volume header present
[WARN ]  EFIVARS: Failed to validate firmware header
[WARN ]  EFIVARS: No Firmware Volume header present
[WARN ]  EFIVARS: Failed to validate firmware header


[NOTE ]  coreboot-dell_optiplex_7010_9010_v0.1.2-rc1-133-gaa3639ed8cec-dirty-v1.0.0-rc3 Sun Jul 06 13:45:35 UTC 2025 x86_32 smm starting (log level: 0)...
[DEBUG]  SMI_STS: APM 
[DEBUG]  Raw write to SMM store, param = 0x493d6000
[DEBUG]  FMAP: area SMMSTORE found @ 1020000 (262144 bytes)
[DEBUG]  smm store: writing 0x49824000 block 0, offset=0x0, size=10000
```

</details>

SergiiDmytruk:
> Not much useful info could be obtained from capsule update process, it just hanged somewhere inside SMI handler, after printing input parameters, already on the first read.

Is that a side-effect of debugging SMM?  Previously Novacustom laptops hanged only with debug changes to IFD/FMAP.

As a workaround I was thinking of changing capsule update in EDK to read FMAP first and then ignore flash read failures unless they touch SI_BIOS/GBE/etc. But if hangs are common for some reason now, this won't work.

krystian-hebel:
> Is that a side-effect of debugging SMM?

I hope not, but at this point it is possible. It doesn't seem to return to UEFI, last printed line was `FMAP: area SMMSTORE found` (or the one before that? can't remember), and not `smm store: reading ...`.

I think I'll just write an app for sending raw SMMSTORE commands, maybe it will at least make reproduction easier so that we won't have to flash new image every time.

SergiiDmytruk:
> I think I'll just write an app for sending raw SMMSTORE commands, maybe it will at least make reproduction easier so that we won't have to flash new image every time.

I just make the update code fail after reading the firmware to not worry about the update, but maybe a separate app won't trash cbmem log buffer.

krystian-hebel:
I was able to generate SMI from Linux userspace, and first observation is that it isn't TCO -> TCO -> APM, but APM -> TCO -> TCO, and TCO isn't generated for APM other than SMMSTORE. It may be possible that TCO bit is set even if SPI transaction is initiated from within SMM.

krystian-hebel:
Tool for doing raw SMMSTORE operations: https://github.com/Dasharo/coreboot/pull/713

In order to use it with full flash access, we would have to disable capsule handling in edk2.

filipleple:
@SergiiDmytruk  so it looks like this branch fixes the capsule update issues. I've successfully tested on V540TU, V560TU, V540TNX - testing on V560TNX is in progress. Please rebase and fix the commit message names, so we can get to merging this

filipleple (dismissed):
also fixed [SMM BWP on NUC BOX](https://github.com/Dasharo/dasharo-issues/issues/1349) :fire:

src/soc/intel/common/block/fast_spi/mmap_boot.c
```diff
@@ -22,28 +22,47 @@ enum window_type {
...
-			  flash_base, size);
+	mem_region_device_ro_init(&mmap_shadow_devs[type], (void *)host_base, size);
+	xlate_window_init(&mmap_windows[type], &mmap_shadow_devs[type].rdev, flash_base, size);
 	printk(BIOS_INFO, "%s: ",
 		(type == FIXED_DECODE_WINDOW) ?
 		 "Fixed Decode Window" : "Extended Decode Window");
 	printk(BIOS_INFO, "SPI flash base=0x%lx, Host base=0x%lx, Size=0x%zx\n",
 	       flash_base, host_base, size);
+
+	real_dev_windows[real_dev_win_count++] = mmap_windows[type];
```
krystian-hebel: I would add a check against out-of-bounds writes, here and below. It will be easier in the future to see an error than accidentally overwriting data in neighboring variable, if it will ever be updated again.

In general, IMHO mixing iterated array indices and predefined positions is messy, but given how it is passed to `xlate_region_device_ro_init()` leaves us no choice, other than reworking the whole driver, which is beyond the scope of the task.

src/soc/intel/common/block/fast_spi/mmap_boot.c
```diff
@@ -22,28 +22,47 @@ enum window_type {
...
+}
+
+static void initialize_rw_window(uintptr_t flash_base, size_t size)
+{
+	const struct region_device *rw_device = boot_device_rw();
+
+	xlate_window_init(&rw_window, rw_device, flash_base, size);
+	printk(BIOS_INFO, "Fast SPI RW Decode Window: SPI flash base=0x%zx, Size=0x%zx\n",
```
krystian-hebel:
```suggestion
	printk(BIOS_INFO, "Fast SPI RW Decode Window: SPI flash base=0x%lx, Size=0x%zx\n",
```

krystian-hebel (changes requested):

src/drivers/efi/capsules.c
```diff
@@ -792,8 +793,20 @@ static void enable_capsule_smi(void *unused)
 {
 	uint32_t ret;
 
-	ret = call_smm(APM_CNT_SMMSTORE, SMMSTORE_CMD_USE_FULL_FLASH,
-		       (void*)(uintptr_t)uefi_capsule_count);
+	/* SMM can occasionally ignore SMI calls, so retry several times on failure. */
```
krystian-hebel: Nit: technically, SMM can't ignore SMI because CPU is not in SMM at that point.

payloads/external/edk2/Kconfig.dasharo
```diff
@@ -4,7 +4,7 @@ config EDK2_REPOSITORY
 	default "https://github.com/Dasharo/edk2"
 
 config EDK2_TAG_OR_REV
-	default "e4dea91cef0af9dbb05e7a407bb18ca18080cbaf"
+	default "e865b79c4e545c89362330792573fc70861b71d3"
```
krystian-hebel: edk2 will need a rebase, so this will change.

SergiiDmytruk: https://github.com/Dasharo/coreboot/compare/703bb4ead23ab121aafc7c79c3aff118e128116e..e023501f15345e867265d6d9f7b762a4b2d10069#diff-f8ea1bb7fc113c7c1afaafc8cf4917177c1c7adad64c066b424c8fc88e88948f

src/drivers/efi/capsules.c
SergiiDmytruk: https://github.com/Dasharo/coreboot/compare/703bb4ead23ab121aafc7c79c3aff118e128116e..e023501f15345e867265d6d9f7b762a4b2d10069#diff-16e632f09b8efab02dabf756cf47fe4e2b67fed218a821a178270840e5900a1b

src/soc/intel/common/block/fast_spi/mmap_boot.c
SergiiDmytruk: https://github.com/Dasharo/coreboot/compare/703bb4ead23ab121aafc7c79c3aff118e128116e..e023501f15345e867265d6d9f7b762a4b2d10069

src/soc/intel/common/block/fast_spi/mmap_boot.c
SergiiDmytruk: https://github.com/Dasharo/coreboot/compare/703bb4ead23ab121aafc7c79c3aff118e128116e..e023501f15345e867265d6d9f7b762a4b2d10069#diff-f647b77e81916109378b68f46e5d9dfe122031abc093d31272b18a18dd074f32

krystian-hebel (approved):

PR #701, opened 2025-06-23
Script for creating a tag with release notes in the message

krystian-hebel:
Example usage:

```bash
./util/dasharo/tag-with-changelog qemu_q35_v0.2.1 qemu_q35_v0.2.2-rc1
```

TBD: this doesn't include changes in the payload, except if explicitly mentioned in changes to config files.


krystian-hebel:
Sometimes two lines are merged into one, I haven't looked as to why yet. It may be related to the fact that I'm currently using the script for non-linear history between two tags (`protectli_vault_jsl_v0.9.3` -> `protectli_vault_jsl_v0.9.4-rc1` + this PR cherry-picked on top). Example:

```
#
##### Intel Braswell (4 commits) #####
- device/device.h: Rename pci_domain_scan_bus- soc/intel/braswell/gpio_support: drop unused get_gpio
- soc/intel/braswell/gpio_support: drop unused get_gpio
- soc/intel/braswell/Kconfig: Set HPET_MIN_TICKS
- soc/intel/braswell: Use {read,write}32p()
```

It seems that there is roughly one such case per group, but it isn't always the first one. The second part is a repetition of another bullet, but is isn't always the immediate one, at least not after sorting.

util/dasharo/tag-with-changelog
```diff
@@ -45,35 +40,38 @@ else
...
 		exit 1
 	fi
+
+	# Create a temporary lightweight tag, exit if it exists already
+	echo "Creating temporary $NEW_GIT_VERSION tag"
+	git tag $NEW_GIT_VERSION || exit 1
+
+	# Try to verify that the repo is clean before losing state.
+	if ! git diff-index --quiet --cached HEAD 2>/dev/null || \
+		[ "$(git diff $NEW_GIT_VERSION --shortstat 2>/dev/null | tail -n1)" != "" ]; then
```
SergiiDmytruk: Just
```suggestion
	if ! git diff --quiet HEAD; then
```
instead of that upstream horror.

util/dasharo/tag-with-changelog
```diff
@@ -45,35 +40,38 @@ else
...
+		echo "ERROR: repo is not clean.  Exiting." >&2
+		echo "Deleting temporary tag"
+		git tag -d $NEW_GIT_VERSION
```
SergiiDmytruk: With the change above adding a tag and then removing it is unnecessary, just make the tag after the check.

util/dasharo/tag-with-changelog
```diff
@@ -396,14 +402,18 @@ for module in $(git submodule --quiet foreach pwd); do
...
-# Add the collected data to the top of the existing logfile for parsing
-if [ -n "$UPDATE_MAIN_LOGFILE" ]; then
-	tmpfile="$(mktemp "LOGFILE.XXXX")"
-	grep -Fxv -f "$MAIN_LOGFILE" "$LOGFILE" > "$tmpfile"
-	printf "\n\n" >> "$tmpfile"
-	cat "$MAIN_LOGFILE" >> "$tmpfile"
-	mv "$tmpfile" "$MAIN_LOGFILE"
-	rm -f "$LOGFILE"
+# Create signed tag, replace existing, use LOGFILE as message, open editor
+git tag -s -f -F $LOGFILE -e $NEW_GIT_VERSION
```
SergiiDmytruk: I'd add quotes around `$LOGFILE` here and below. Version is unlikely to need escaping, but a full path could.  Although, I'd add quotes around `$NEW_GIT_VERSION` too, just because if it's empty or weird in some way, the errors will be more intelligible.

util/dasharo/tag-with-changelog
```diff
@@ -0,0 +1,426 @@
...
+	# Leave ${paths} unquoted in the git commands
+	# shellcheck disable=SC2086
+	{ \
+		if [ -n "$paths" ]; then \
+			git log --abbrev-commit --pretty=format:"- %s" \
+				"${oldver}..${newver}" -- ${paths} \
+				2>/dev/null; \
+		fi; \
+		if [ -n "$keywords" ]; then \
+			git log --abbrev-commit --pretty=format:"- %s" \
```
SergiiDmytruk:
```suggestion
			git log --abbrev-commit --pretty=tformat:"- %s" \
```
And above as well.  It makes git add newline character even after printing a single line of output, otherwise some lines get glued as you have already noticed.

SergiiDmytruk (changes requested):

util/dasharo/tag-with-changelog
krystian-hebel: It was added as part of https://review.coreboot.org/c/coreboot/+/62440, I suspect it was added to `Handle symbolic links to the git repo`. We probably don't have to deal with this corner case, though, it may be required for some specific downstream build system.

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/74d14b8e589dc17fe4381940f7df82be6c8ce81b..4c3f9dc7da460387649522b54bcb68cf2f6a9f31

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/74d14b8e589dc17fe4381940f7df82be6c8ce81b..4c3f9dc7da460387649522b54bcb68cf2f6a9f31

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/74d14b8e589dc17fe4381940f7df82be6c8ce81b..4c3f9dc7da460387649522b54bcb68cf2f6a9f31

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/74d14b8e589dc17fe4381940f7df82be6c8ce81b..4c3f9dc7da460387649522b54bcb68cf2f6a9f31

SergiiDmytruk: > I suspect it was added to `Handle symbolic links to the git repo`

I don't see how, but see 2 new invocations of `realpath` at the bottom which are more likely candidates for handling symlinks.

`git diff-index --quiet --cached HEAD 2>/dev/null` checks for empty index (so staged changes) and `[ "$(git diff $NEW_GIT_VERSION --shortstat 2>/dev/null | tail -n1)" != "" ]` looks like a terrible way for checking for unstaged changes.  Probably a case of unfortunate adjustment while making a release.

util/dasharo/tag-with-changelog
```diff
@@ -166,7 +166,7 @@ get_log_submodule() {
 # up manually.
 version_ctrl_c() {
 	printf "\n Cleaning up and exiting.\n" >&2
-	git checkout origin/main > /dev/null 2>&1
+	git checkout $NEW_GIT_VERSION > /dev/null 2>&1
```
SergiiDmytruk: Then this can also be quoted.

util/dasharo/tag-with-changelog
```diff
@@ -200,7 +200,7 @@ for module in $(git submodule --quiet foreach pwd); do
 	declare "$version"="$(get_latest_commit_id "$module")"
 done
 
-git checkout origin/main  > /dev/null 2>&1
+git checkout $NEW_GIT_VERSION  > /dev/null 2>&1
```
SergiiDmytruk: As well as this can be quoted.  This makes `shellcheck` happy (there are 2 stylistic notes unrelated to the changes).

util/dasharo/tag-with-changelog
```diff
@@ -18,62 +18,62 @@ then
...
-fi
-
 # Verify the command line arguments
 if [ "$1" == "--help" ] || [ -z "$1" ] || [ -z "$2" ]; then
 	echo
 	echo "Usage: $0 <old_version> <new_version> [release notes file]"
 	echo "Old version should be a tag (4.1), a branch (origin/4.1), or a commit id"
-	echo "New version can be a branch (origin/main) a tag (4.2), or a commit id"
-	echo "Logfile can be a new file or an existing file to update"
+	echo "New version must be a name of a new tag, this script fails if it exists"
```
SergiiDmytruk, line 26: The script no longer fails if the new tag exists because that was a side-effect of creating the tag.  This text or the code needs to change.

SergiiDmytruk (changes requested):

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/b20dcc26401c9ec1656c191cc2d9e1c47ab0f743..97105c6159c8ca5a10eccf425310b9f4339ad377

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/b20dcc26401c9ec1656c191cc2d9e1c47ab0f743..97105c6159c8ca5a10eccf425310b9f4339ad377

util/dasharo/tag-with-changelog
krystian-hebel: I've decided to change the code to fail in that case, I think it would be safer than accidentally modifying the tags.

https://github.com/Dasharo/coreboot/compare/b20dcc26401c9ec1656c191cc2d9e1c47ab0f743..97105c6159c8ca5a10eccf425310b9f4339ad377

SergiiDmytruk (dismissed):
LGTM.  Not merging in case you decide to do further changes here.

krystian-hebel:
I've added edk2 changelog generation, although it will be working best two releases from now, as there is no easy way of getting common history from multiple defconfigs without somehow parsing tag name and mapping it to config name. Now edk2 revision is common between all Dasharo platforms. In theory it should still be possible to specify different one in config file, but that would once again make things harder to follow.

I don't plan to add anything more in this PR. One possible improvement could be reusing repo in `payloads` instead of downloading a fresh one, but:
- tags aren't created that often so downloading another copy every now and than shouldn't be too annoying
- payloads directories are listed in `.gitignore`, meaning their contents aren't protected by check for clean main repo

util/dasharo/tag-with-changelog
```diff
@@ -410,6 +422,26 @@ for module in $(git submodule --quiet foreach pwd); do
...
+# isn't trivial to differentiate between such case and use as placeholder value.
+PREHISTORY="f0459dff1425c0c02d0736de5737eb23a1b671ec"
+
+if [ -z $OLD_EDK2_GIT_VERSION ]; then
+	echo "# Old EDK2 version couldn't be obtained, changelog may be incomplete!" >> "$LOGFILE"
+	echo "# Commits older than ${PREHISTORY:0:10} won't be shown" >> "$LOGFILE"
+	OLD_EDK2_GIT_VERSION="$PREHISTORY"
+fi
+
+git clone "$URL" || version_ctrl_c
```
SergiiDmytruk: This should be almost as good as reusing that repository when it exists:
```suggestion
git clone --reference-if-able payloads/external/edk2/workspace/Dasharo/ --bare "$URL" edk2 || version_ctrl_c
```
If that path doesn't exist, the repo gets cloned as usual.  `edk2` at the end is needed because `--bare` results in `edk2.git` name being used by default.

util/dasharo/tag-with-changelog
```diff
@@ -410,6 +422,26 @@ for module in $(git submodule --quiet foreach pwd); do
...
+echo "#### EDK2 ####" >> "$LOGFILE"
+echo "##############" >> "$LOGFILE"
+# Hardcoded, probably won't ever change
+URL="https://github.com/Dasharo/edk2"
+# This is the first digest from common Kconfig option. While it could be pointed
+# to by previous release (in which case the changelog would be complete), it
+# isn't trivial to differentiate between such case and use as placeholder value.
+PREHISTORY="f0459dff1425c0c02d0736de5737eb23a1b671ec"
+
+if [ -z $OLD_EDK2_GIT_VERSION ]; then
```
SergiiDmytruk:
```suggestion
if [ -z "$OLD_EDK2_GIT_VERSION" ]; then
```

SergiiDmytruk (commented):
> I've added edk2 changelog generation, although it will be working best two releases from now, as there is no easy way of getting common history from multiple defconfigs without somehow parsing tag name and mapping it to config name.

Name of the defconfig file could be provided on the command-line and `git show "old-revision:configs/$config"` could be used to obtain it, but maybe we don't need to complicate this further given the absence of detailed notes until now.

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/6116e4b5e35d9173a29ac3fa1e4932ff8b6e60f0..0b42ba7fcb2c81fe05999cd29fd4d93c3cce8555#diff-0ecf7672503aa0c7446284c3c0b00569b847900b9e4e0db39822bf42fb7ba62fR435

util/dasharo/tag-with-changelog
krystian-hebel: https://github.com/Dasharo/coreboot/compare/6116e4b5e35d9173a29ac3fa1e4932ff8b6e60f0..0b42ba7fcb2c81fe05999cd29fd4d93c3cce8555#diff-0ecf7672503aa0c7446284c3c0b00569b847900b9e4e0db39822bf42fb7ba62fR441

It prints a warning when the directory doesn't exist, but since it prints download progress anyway, it may serve as an information as to why the download is required in the first place.

krystian-hebel:
> Name of the defconfig file could be provided on the command-line and `git show "old-revision:configs/$config"` could be used to obtain it, but maybe we don't need to complicate this further given the absence of detailed notes until now.

This would be just another place to make a mistake. I asked internally about possibility to modify `build.sh` to make config filename and ROM filename consistent (with hope that tag name could be made consistent as well), but I didn't get any answer. The idea would be to add an option to print that data in machine-readable form that could be consumed e.g. by this script.

SergiiDmytruk (approved):

PR #703, opened 2025-06-24
Fix capsule flashing again

SergiiDmytruk:
Recent https://github.com/Dasharo/coreboot/pull/686 caused trouble (primarily) for chips less than 32 MiB while fixing flash access beyond BIOS region.  We also need to take FUM into account for capsules, so they can write anywhere (similar change in EDK: https://github.com/Dasharo/edk2/pull/257).

*ref: prot-1829*


krystian-hebel (dismissed):
Not merging yet, I have to review edk2 part first.

SergiiDmytruk:
Updated with fixed FastSPI changes, both of which were wrong and effectively undid https://github.com/Dasharo/coreboot/pull/686.

krystian-hebel (approved):

PR #705, opened 2025-06-27
Sv boot integration

miczyg1:


miczyg1:
Hmm the tests on QEMU won't pass. The presence of Sovereign Boto changes the boot flow. THe wizard always coms first, prompt is not printed in this case.

macpijan:
I am glad that we have added this check, and are aware of this prior merging, not after.

Sounds like we need to adjust the tests first prior merging this?

miczyg1:
> I am glad that we have added this check, and are aware of this prior merging, not after.
> 
> Sounds like we need to adjust the tests first prior merging this?

I don't think any test adjustment will work here. The boot flow looks much different than usual. I think the best would be to have a separate QEMU config with the wizard enabled, because when the wizard is enabled, only the tests from the [new suite](https://github.com/Dasharo/open-source-firmware-validation/pull/915) will work

PR #709, opened 2025-07-03
mb/protectli/vault_jsl/devicetree.cb: disable wake on USB attach

mkopec:
Restores v0.9.3 behavior of not waking up instantly after entering S3. It seems that the onboard USB3 hub would wake up the platform if wake on attach/detach is enabled.

Upstream-Status: Pending
Change-Id: I6899689ef6814cf46d4808e96efa7c232877fa3d

### DIFF
--- a/src/mainboard/protectli/vault_jsl/devicetree.cb
+++ b/src/mainboard/protectli/vault_jsl/devicetree.cb
@@ -41,8 +41,8 @@ chip soc/intel/jasperlake
 		[5] = USB3_PORT_DEFAULT(OC_SKIP),	// M.2 LTE
 	}"
 
-	register "usb2_wake_enable_bitmap" = "0x0012"
-	register "usb3_wake_enable_bitmap" = "0x0003"
+	register "usb2_wake_enable_bitmap" = "0x0002"
+	register "usb3_wake_enable_bitmap" = "0x0001"
 
 	register "SataSalpSupport" = "1"
 	register "SataPortsEnable[1]" = "1"


krystian-hebel:
https://github.com/Dasharo/coreboot/blob/protectli_vault_jsl_v0.9.3/src/mainboard/protectli/vault_jsl/devicetree.cb#L47 - those lines were present in v0.9.3, yet the sleep works there. Has something changed elsewhere that broke it?

mkopec:
yes, https://review.coreboot.org/c/coreboot/+/78086 was merged

miczyg1:
> https://github.com/Dasharo/coreboot/blob/protectli_vault_jsl_v0.9.3/src/mainboard/protectli/vault_jsl/devicetree.cb#L47 - those lines were present in v0.9.3, yet the sleep works there. Has something changed elsewhere that broke it?

It must be something else breaking it. The USB ports on JSL boards are powered in S3 and even S5, because power in USB connectors is supplied from +V5A and controlled via SIO GPIO: https://github.com/Dasharo/coreboot/blob/protectli_vault_jsl_v0.9.3/src/mainboard/protectli/vault_jsl/bootblock.c#L35

The wake via USB is supposed to work here.

mkopec:
@miczyg1 `usb2_wake_enable_bitmap` controls wake on usb *attach/detach* . Not wake by USB keyboard keypress. 2 different things.

Now as for wake by USB keyboard keypress, it still works after this change **but only on the one USB-A port that is wired directly from the PCH**, not from the hub. The hub ports do not support wake by usb keyboard.

miczyg1:
> @miczyg1 `usb2_wake_enable_bitmap` controls wake on usb _attach/detach_ . Not wake by USB keyboard. 2 different things.

Yes, I know what it does.

> Now as for wake by USB keyboard, it still works after this change **but only on the one USB-A port that is wired directly from the PCH**, not from the hub. The hub ports do not support wake by usb keyboard.

Does it mean that we have to remove wake on attach/detach from all USB ports?

mkopec:
> Does it mean that we have to remove wake on attach/detach from all USB ports?

Theoretically we can enable it on the 1 special PCH USB port and it should work, but then one port would work differently from the rest, and I'm not sure if the client would be happy with that

miczyg1:
> > Does it mean that we have to remove wake on attach/detach from all USB ports?
> 
> Theoretically we can enable it on the 1 special PCH USB port and it should work, but then one port would work differently from the rest, and I'm not sure if the client would be happy with that

Still better than nothing. We can disable it if it is requested,

Found the hub datasheet, it may help: https://datasheet.lcsc.com/lcsc/1808111624_VIA-Tech-VL817-Q7-B0_C209756.pdf

I see it is powered with standby 5V rather than V5A, so hub is not powered. So all we need is to disable the bitmaps for USB3.0 and 2.0 ports connected to the hub.

mkopec:
> I see it is powered with standby 5V rather than V5A, so hub is not powered. So all we need is to disable the bitmaps for USB3.0 and 2.0 ports connected to the hub.

OK, we can do that: https://github.com/Dasharo/coreboot/pull/709/commits/06204e16585cb02f5a0041ab5f64c2715ebd9d4b

SergiiDmytruk:
I think the commit message should capture comments made above on the effect of https://review.coreboot.org/c/coreboot/+/78086, attach/detach, pch/hub.

mkopec:
Updated the commit message @SergiiDmytruk

SergiiDmytruk (approved):

PR #710, opened 2025-07-07
Nuc vboot fix

mkopec:
Vboot works, but the SMM BWP option still causes issues (boot hang, options resetting, failure submitting form in setup menu)


mkopec:
There seems to be instability in SMMSTORE. Following @SergiiDmytruk 's fix here: https://github.com/Dasharo/edk2/pull/255/commits/d69448102e9f951e1669018a614099f65554613f I decided to bump the number of retries even further to 100 and it seems to have mitigated this issue, but I don't think that's the proper solution...

PR #711, opened 2025-07-07
payloads/Makefile.mk: add new option to skip auto flash

wiktormowinski:
solution to: https://github.com/Dasharo/dasharo-issues/issues/1402

edk2 part: https://github.com/Dasharo/edk2/pull/264


payloads/external/edk2/Kconfig.dasharo
```diff
@@ -413,4 +413,11 @@ config EDK2_GRAPHICAL_CAPSULE_PROGRESS
 	  An alternative is a textual one which also appears on serial, the default
 	  is graphical.
 
+config CONFIG_EDK2_DO_NOT_BOOT_DTS_IN_FUM
+	bool "Don't launch the auto-update sequence in Firmware Update Mode"
+	default n
+	help
+	  A way to disable all flash locks on and use the platform manually, the default
```
SergiiDmytruk: What this the intention?
```suggestion
	  A way to disable all flash locks on the next boot and boot the platform as usual, the default
```

SergiiDmytruk (commented):
How about a name like `FUM_AUTO_IPXE_BOOT`?  EDK doesn't know anything about DTS, it's baked into iPXE image by coreboot which then supplies it to EDK as part of the build.  Also it's better to avoid negation in names as they lead to double negations in conditionals.

SergiiDmytruk:
And one more thing: the new option should be utilized in `configs/config.protectli_vp66xx`.

payloads/external/edk2/Kconfig.dasharo
wiktormowinski: sure, im not good at naming things: 
https://github.com/Dasharo/coreboot/pull/711/commits/24896daeaceb8dc0dd94056757a365ca9c9d3b5a

SergiiDmytruk:
Changing the name inverted meaning of the option and `config` was missing `EDK2` prefix, but I corrected that while testing the functionality.

Also note that you weren't using latest `dasharo` branch locally.  EDK's revision is no longer specified in `configs`, see https://github.com/Dasharo/coreboot/pull/711/commits/6b6440ac4fb86cbe1a14b9d33e689a2b27d456df for the new way of doing it.

SergiiDmytruk (approved):

PR #718, opened 2025-07-21
configs/config.novacustom_nuc_box: disable ccache

mkopec:
Fixes building with dasharo-sdk and aligns the config with other platforms' configs.

Change-Id: I85d26f60426c05d7aaff92dd74bd34cb5ac06a72


matmacieje (approved):
Works for me.

PR #725, opened 2025-07-25
Make capsule_cabinet work independently of buildsystem state

mkopec:
Extract the config from the capsule itself, to avoid depending on .config existing and being up to date.

This depends on edk2 basetools, these can be taken from the edk2 build tree, or specified as an env variable.


mkopec:
Draft, because scripts still depend on /payloads/external/edk2/workspace being populated

filipleple (dismissed):

capsule_cabinet.sh
```diff
@@ -24,17 +27,6 @@ if [ ! -f .config ]; then
     die "No '.config' file in current directory"
 fi
```
SergiiDmytruk, line 28: Existence of `.config` is checked here.

capsule_cabinet.sh
```diff
@@ -24,17 +27,6 @@ if [ ! -f .config ]; then
...
-    fi
-done <<< "$(sed 's/\$(\([^)]\+\))/${\1}/g' .config)"
-
-if [ "$CONFIG_DRIVERS_EFI_UPDATE_CAPSULES" != y ]; then
-    die "Current board configuration lacks support of update capsules"
-fi
-
 capsule=$1
 date=$(stat -c %w $capsule | cut -d ' ' -f 1)
 vendor=$(cat .config | grep -e "CONFIG_VENDOR_.*=y" | cut -d '=' -f 1 | cut -d '_' -f 3- | awk '{ print tolower($0) }')
```
SergiiDmytruk, line 32: `.config` is used here.

SergiiDmytruk (changes requested):
Unless I misunderstand the intend, the changes are incomplete.  The script can't be used without `.config` in the current directory and it will mix configuration from `.config` and coreboot's config extracted from a capsule's payload.

PR #750, opened 2025-08-22
build.sh: sdk_run: Use `:Z` to relabel volumes attached to docker

philipanda:
When the repository is cloned to a directory with a domain that is not accessible by containers by default (like user_home_t), the build might fail due to access denials, or just fail silently. 

`ausearch -m avc -ts recent -i` shows violation logs like:
```
type=AVC msg=audit(08/22/2025 09:43:42.740:641) : avc: denied { write } for pid=18611
comm=rm name=coreboot dev="dm-0" ino=85471
scontext=system_u:system_r:container_t:s0:c100,c319
tcontext=unconfined_u:object_r:user_home_t:s0
tclass=dir permissive=1
```
- `container_t` cannot access `user_home_t`, so the access is denied.

Adding `:Z` suffix to the volumes relabels them with a domain that is accessible to `container_t` in SELinux. Specifically `:Z` says that the volumes will be only accessible by this one single container, and not shared between multiple ones. This way the build.sh script can be used without issues on SELinux-enabled systems without changing SELinux enforcement mode.
Source: https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label


pietrushnic (dismissed):
Woeks for me

philipanda:
@pietrushnic does your system use SELinux? I wonder how this affects systems without it

pietrushnic:
> @pietrushnic does your system use SELinux? I wonder how this affects systems without it

I don't think so, I didn't enable it on my Debian 12. `sestatus` is not installed, and I guess that would be the way to check. I just checked if that change will cause issues for me.

PR #759, opened 2025-08-31
TPM fixes for Jasper Lake

SergiiDmytruk:
* Reserve TPM's MMIO range (wasn't causing problems, but it should be reserved).
* Bump EDK for a workaround there.

EDK PR: https://github.com/Dasharo/edk2/pull/273
*ref: prot-1876*

### DIFF
--- a/payloads/external/edk2/Kconfig.dasharo
+++ b/payloads/external/edk2/Kconfig.dasharo
@@ -4,7 +4,7 @@ config EDK2_REPOSITORY
 	default "https://github.com/Dasharo/edk2"
 
 config EDK2_TAG_OR_REV
-	default "e4616bfb0b40b6ac710fdb2e20110de874cc17c6"
+	default "f363c3ad2b8492f8f357a1c3578191f52395c480"
 
 config EDK2_SYSTEM76_EC_LOGGING
 	bool "Enable edk2 logging to System76 EC"
--- a/src/drivers/crb/tis.c
+++ b/src/drivers/crb/tis.c
@@ -105,6 +105,12 @@ static void crb_tpm_fill_ssdt(const struct device *dev)
 	acpigen_pop_len(); /* Device */
 }
 
+static void crb_tpm_read_resources(struct device *dev)
+{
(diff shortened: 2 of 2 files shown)


src/soc/intel/jasperlake/systemagent.c
```diff
@@ -29,6 +29,9 @@ void soc_add_fixed_mmio_resources(struct device *dev, int *index)
 		{ EPBAR, EP_BASE_ADDRESS, EP_BASE_SIZE, "EPBAR" },
 		{ REGBAR, REG_BASE_ADDRESS, REG_BASE_SIZE, "REGBAR" },
 		{ EDRAMBAR, EDRAM_BASE_ADDRESS, EDRAM_BASE_SIZE, "EDRAMBAR" },
+
+		/* first field (sa_mmio_descriptor.index) is not used, setting to 0: */
+		{ 0, TPM_BASE_ADDRESS, TPM_SIZE, "TPM" },
```
miczyg1: I think it should not be done that way. take a look at drivers/pc80/tpm/tis.c:

```c
static void lpc_tpm_read_resources(struct device *dev)
{
	/* Static 5K memory region specified in Kconfig */
	mmio_range(dev, 0, CONFIG_TPM_TIS_BASE_ADDRESS, 0x5000);
}
```

I would check if adding similar thing to drivers/crb/tis.c here:

```c
static struct device_operations __maybe_unused crb_ops = {
	.read_resources = noop_read_resources,
```

would do the trick.

TBH not sure why alderlake is the only one defining TPM range in SA.

miczyg1 (changes requested):

src/soc/intel/jasperlake/systemagent.c
SergiiDmytruk: Updated: https://github.com/Dasharo/coreboot/compare/629e34792aed2290429781ee57160132adc1cbbd..26c73446d67c9ec9e6b6d5a48defc168f503b998

Can see the range being reserved on VP1210.

miczyg1 (approved):

PR #760, opened 2025-09-01
vc/dasharo/options.h: Expose descriptor writeability flag as EFI var

mkopec:
Also see https://github.com/Dasharo/edk2/pull/274

### DIFF
--- a/src/vendorcode/dasharo/options.c
+++ b/src/vendorcode/dasharo/options.c
@@ -273,8 +273,12 @@ uint8_t cse_get_me_disable_mode(void)
 		read_u8_var("MeMode", &var);
 	}
 
-	/* Disable ME via HMRPFO if in Firmware Update Mode */
-	if (CONFIG(DASHARO_FIRMWARE_UPDATE_MODE) && fum) {
+	/*
+	 * Disable ME via HMRPFO if in Firmware Update Mode
+	 * Don't do it if capsules are supported, as capsule updates are not
+	 * currently compatible with HMRFPO
+	 */
+	if (CONFIG(DASHARO_FIRMWARE_UPDATE_MODE) && fum	&& !(CONFIG(DRIVERS_EFI_UPDATE_CAPSULES))) {
 		/* Check if already in HMRFPO mode */
 		if (cse_is_hfs1_com_secover_mei_msg())
 			return ME_MODE_DISABLE_HMRFPO;
--- a/src/vendorcode/dasharo/options.c
+++ b/src/vendorcode/dasharo/options.c
@@ -1,8 +1,11 @@
 /* SPDX-License-Identifier: GPL-2.0-only */
 
+#include <bootstate.h>
 #include <console/console.h>
 #include <dasharo/options.h>
+#include <device/mmio.h>
 #include <drivers/efi/efivars.h>
+#include <fmap.h>
 #include <option.h>
 #include <soc/intel/common/reset.h>
 #include <smmstore.h>
@@ -285,7 +288,7 @@ uint8_t cse_get_me_disable_mode(void)
 enum dgpu_state dasharo_dgpu_state(void)
 {
 	uint8_t dgpu_state = NVIDIA_OPTIMUS;
-	/* 
+	/*
 	 * 0 - IGPU_ONLY
 	 * 1 - NVIDIA_OPTIMUS (iGPU+dGPU)
 	 * 2 - DGPU_ONLY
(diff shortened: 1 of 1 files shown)


src/vendorcode/dasharo/options.c
```diff
@@ -540,3 +543,68 @@ bool get_ibecc_option(bool ibecc_default)
...
+	if (rdev_readat(&desc_rdev, si_desc_buf, 0, SI_DESC_SIZE) != SI_DESC_SIZE) {
+		printk(BIOS_ERR, "Failed to read Descriptor Region from SPI Flash\n");
+		return;
+	}
+
+	descriptor_writeable = is_descriptor_writeable(si_desc_buf);
+
+	printk(BIOS_DEBUG, "Descriptor is %swriteable\n", descriptor_writeable ? "" : "not ");
+
+	efi_fv_set_option(&smmstore_rdev,
```
miczyg1, line 608: I recall writing a variable was not yet supproted in the initial patchset of SMMSTORE EFI variables. I think you may have some problems here I, at least, had a lot if variables were READ multiple times in one stage, because rdev pointer was being overwritten and ended up at the end of storage when attempted to read variable 2nd time or more.

SergiiDmytruk: We have your fix in https://github.com/Dasharo/coreboot/commit/788c03436b15, but I remember using it and still having trouble when trying to write XMP profile to an EFI variable.  It's possible that this write won't be affected by such issues.

By the way, coreboot never creates variable store, so nothing can be communicated via it on the first boot or if the store needs recovery.  I guess IFD's lock state is not exposed in any register?  We can't read IFD from EDK because SmmStore allows that only when capsules are present.

mkopec: IFD lock can be read from the descriptor itself (which would require reading and parsing the descriptor in edk2) or from TGL onwards through HFSTS1 bit 4 (which isn't available while ME is disabled, unless https://github.com/Dasharo/coreboot/pull/453 is used)

mkopec:
V540TU Capsule Update now passes. Needed to remove the HMRFPO disabling code (uncommented, to allow easy re-adding once needed again)

src/vendorcode/dasharo/options.c
```diff
@@ -271,21 +274,21 @@ uint8_t cse_get_me_disable_mode(void)
...
 
 	/* Disable ME via HMRPFO if in Firmware Update Mode */
-	if (CONFIG(DASHARO_FIRMWARE_UPDATE_MODE) && fum) {
-		/* Check if already in HMRFPO mode */
-		if (cse_is_hfs1_com_secover_mei_msg())
-			return ME_MODE_DISABLE_HMRFPO;
-		else
-			disable_me_hmrfpo(&var);
-	}
+	//if (CONFIG(DASHARO_FIRMWARE_UPDATE_MODE) && fum) {
```
SergiiDmytruk: That's quite radical, could have just added `CONFIG(DRIVERS_EFI_UPDATE_CAPSULES)` and a comment that locked IFD implies no HAP during capsule updates and we must avoid cold reset on enabling HMRFPO.  The check could even be more fine grained and additionally check for IFD being locked.  I just worry of some unexpected change in behaviour during updates on other platforms.

mkopec: Changed to disable HMRFPO if capsule update is enabled https://github.com/Dasharo/coreboot/pull/760/commits/c50cf495a488d2d1736a60cdd32198cc25579a7d
Decided not to include the IFD check as this issue affects all platforms using capsule update, not only the ones with a locked descriptor.

src/vendorcode/dasharo/options.c
mkopec: @miczyg1 can this thread be resolved?

mkopec:
@SergiiDmytruk is there anything else missing here?

src/vendorcode/dasharo/options.c
SergiiDmytruk: We'll probably want to pass this data via cbmem in the future to have first boot behave the same as any other, but this should do for now.  If testing won't demonstrate any breakage, adding variables at the end of the ramstage may be working fine.  I was trying to do it in romstage.

BeataZdunczyk: @SergiiDmytruk So creating a GH issue should be enough for this release, and with the next IBG implementation we can start addressing this issue?

SergiiDmytruk (dismissed):
> @SergiiDmytruk is there anything else missing here?

No (other than CI), I was doing some tests with QEMU to verify UI updates and update blocking, seems to work as expected.

src/vendorcode/dasharo/options.c
SergiiDmytruk: @BeataZdunczyk, yes, it's not a big deal, users probably won't even notice and the condition should be gone after a reboot anyway.  But it already confused me while testing and can also affect automated tests.

mkopec: https://github.com/Dasharo/dasharo-issues/issues/1560

BeataZdunczyk: >  But it already confused me while testing and can also affect automated tests.

Which tests? IBG: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/dasharo-security/cbnt.robot?

miczyg1:
FYI: I already bumped edk2 rev here: https://github.com/Dasharo/coreboot/pull/762 (it is newer than the one updated here)

src/vendorcode/dasharo/options.c
miczyg1: HFSTS1 bit4 may not necessarily show the descriptor lock. E.g. if descriptor is locked, but ME is not, then this bit may be clear...

But I have good news for you. Check `BIOS_FDOC` and `BIOS_FDOD` in the SPI MMIO (PCH EDS). These registers can be used to read the descriptor without using the whole SPI chip reads, just plain and simple index/data pair. I believe it existed since ICH era...

@mkopec

SergiiDmytruk: > Which tests? IBG: https://github.com/Dasharo/open-source-firmware-validation/blob/develop/dasharo-security/cbnt.robot?

@BeataZdunczyk, no, capsule tests and menu tests if a binary with locked IFD is used.  I don't think we do that at the moment.

miczyg1 (approved):

PR #763, opened 2025-09-02
configs/config.hardkernel_odroid_h4: Bump to v0.9.1 suffix

miczyg1:
Upstream-Status: Inappropriate [custom configs]


miczyg1:
@SergiiDmytruk please review if I changed all necessary values correctly so that I can place final tag

configs/config.hardkernel_odroid_h4
```diff
@@ -17,8 +17,8 @@ CONFIG_PCIEXP_SUPPORT_RESIZABLE_BARS=y
 CONFIG_DRIVERS_EFI_VARIABLE_STORE=y
 CONFIG_DRIVERS_EFI_FW_INFO=y
 CONFIG_DRIVERS_EFI_MAIN_FW_GUID="cd18c3e8-b350-4fac-bc1a-36621db26999"
-CONFIG_DRIVERS_EFI_MAIN_FW_VERSION=0x00090103
-CONFIG_DRIVERS_EFI_MAIN_FW_LSV=0x00090103
+CONFIG_DRIVERS_EFI_MAIN_FW_VERSION=0x00090100
```
SergiiDmytruk: Last byte [should be](https://docs.dasharo.com/kb/capsule-updates-configs/) `0x80` for releases in both files.  LSV can end with `00` if you want to allow downgrades to RC versions.  CI catches this too.

SergiiDmytruk (changes requested):

configs/config.hardkernel_odroid_h4
miczyg1: Rather no downgrades. So both should have 0x80?

miczyg1: 6ca33f0e4e8

SergiiDmytruk: > Rather no downgrades. So both should have 0x80?

Yes.  Version of an update is compared simply with `>= CONFIG_DRIVERS_EFI_MAIN_FW_LSV` (effectively) during validation.

SergiiDmytruk (approved):

PR #771, opened 2025-09-11
Add Protectli logo within Make buildsys, don't patch post-build

filipleple:
https://github.com/Dasharo/dasharo-issues/issues/1512

*ref: prot-1894*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,6 +337,23 @@ ifeq ($(FAILBUILD),1)
 $(error cannot continue build)
 endif
 
+# Add bootsplash to BOOTSPLASH/logo.bmp if a file is provided
+ifneq ($(CONFIG_BOOTSPLASH_REGION_LOGO_FILE),"")
+
+files_added:: branding_bootsplash_region
+
+.PHONY: branding_bootsplash_region
+branding_bootsplash_region: $(obj)/coreboot.rom $(objutil)/cbfstool/cbfstool
+	@echo "    BRANDING   Adding bootsplash (BOOTSPLASH/logo.bmp) from $(CONFIG_BOOTSPLASH_REGION_LOGO_FILE)"
+	$(objutil)/cbfstool/cbfstool $(obj)/coreboot.rom add \
+		-r BOOTSPLASH \
+		-f $(CONFIG_BOOTSPLASH_REGION_LOGO_FILE) \
+		-n logo.bmp \
+		-t raw \
+		-c lzma
+
+endif
+
 # Run hooks registered by subdirectories that need to be evaluated after all files have been parsed
 $(eval $(postinclude-hooks))
 
--- a/build.sh
+++ b/build.sh
@@ -146,7 +146,6 @@ function build_msi {
 function build_protectli_vault {
   DEFCONFIG="configs/config.protectli_${BOARD}"
   FW_VERSION=$(cat ${DEFCONFIG} | grep CONFIG_LOCALVERSION | cut -d '=' -f 2 | tr -d '"')
-  LOGO="3rdparty/dasharo-blobs/protectli/bootsplash.bmp"
 
   build_prep
 
@@ -155,14 +154,6 @@ function build_protectli_vault {
   build_start
 
   cp build/coreboot.rom protectli_${BOARD}_${FW_VERSION}.rom
-
-  sdk_run /bin/bash -c "./build/cbfstool \
-    protectli_${BOARD}_${FW_VERSION}.rom add \
-    -r BOOTSPLASH \
-    -f \"$LOGO\" \
-    -n logo.bmp \
-    -t raw \
-    -c lzma"
 
   if [ $? -eq 0 ]; then
     echo "Result binary placed in $PWD/protectli_${BOARD}_${FW_VERSION}.rom"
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -337,6 +337,31 @@ ifeq ($(FAILBUILD),1)
 $(error cannot continue build)
 endif
 
+# Add bootsplash to BOOTSPLASH/logo.bmp on Protectli builds.
+define protectli_branding_hook
+
+# Gate on vendor; use filter to avoid whitespace pitfalls
+ifneq ($(filter y,$(CONFIG_VENDOR_PROTECTLI)),)
```
mkopec: can we make it generic to all platforms by introducing a Kconfig option instead of checking the vendor?

SergiiDmytruk:
Doesn't this effectively undo 831ace521e441bc46c4cb80da35f06cb024007aa by ignoring what's in the defconfig/.config?

Makefile
filipleple: @mkopec hmmm on second thought, wouldn't using the already available `CONFIG_BOOTSPLASH_IMAGE` have the same effect?

filipleple: @mkopec I've [added](https://github.com/Dasharo/coreboot/pull/771/files#diff-e27891368e9cd19dce44ed6d6f56ee8432b019f8eef30b48947df09af9957e9cR477) a Kconfig option and [using it](https://github.com/Dasharo/coreboot/pull/771/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R340) in the Makefile

filipleple:
@SergiiDmytruk Not quite, I think what's key is that there's a black "empty" default image built into the payload, and then there's the customizable BOOTSPLASH region that's set to the Protectli logo. The customer wanted to be able to remove the logo to display the fallback black screen instead. 

Since we always want the logo to be in there in the binaries we deliver to the customer, I think it's more failsafe when it's inserted within the build than when we have to remember to set an extra flag in `dasharo/dasharo` to get the exact binary we want

Makefile
```diff
@@ -337,6 +337,29 @@ ifeq ($(FAILBUILD),1)
...
+.PHONY: branding_bootsplash_region
+branding_bootsplash_region: $(obj)/coreboot.rom $(objutil)/cbfstool/cbfstool
+	@echo "    BRANDING   Adding bootsplash (BOOTSPLASH/logo.bmp) from $(CONFIG_BOOTSPLASH_REGION_LOGO_FILE)"
+	$(objutil)/cbfstool/cbfstool $(obj)/coreboot.rom add \
+		-r BOOTSPLASH \
+		-f $(CONFIG_BOOTSPLASH_REGION_LOGO_FILE) \
+		-n logo.bmp \
+		-t raw \
+		-c lzma
+	@:
```
SergiiDmytruk: Looks unnecessary.
```suggestion
```

Makefile
```diff
@@ -337,6 +337,29 @@ ifeq ($(FAILBUILD),1)
 $(error cannot continue build)
 endif
 
+# Add bootsplash to BOOTSPLASH/logo.bmp
+define bootsplash_region_hook
+
+# Only run if a file is provided
+ifneq ($(CONFIG_BOOTSPLASH_REGION_LOGO_FILE),"")
+
+real-target: branding_bootsplash_region
```
SergiiDmytruk:
```suggestion
files_added:: branding_bootsplash_region
```
Otherwise `BRANDING ...` can be printed after the final `Built ...` line.  This placement is also not ideal:
```
    CBFSPRINT  coreboot.rom

    BRANDING   Adding bootsplash (BOOTSPLASH/logo.bmp) from 3rdparty/dasharo-blobs/protectli/bootsplash.bmp
FMAP REGION: COREBOOT
Name                           Offset     Type           Size   Comp
```
but it's better and I think `files_added` is how such modifications must be done.

Makefile
```diff
@@ -337,6 +337,29 @@ ifeq ($(FAILBUILD),1)
...
+		-f $(CONFIG_BOOTSPLASH_REGION_LOGO_FILE) \
+		-n logo.bmp \
+		-t raw \
+		-c lzma
+	@:
+
+endif
+endef
+
+postinclude-hooks += $(bootsplash_region_hook)
```
SergiiDmytruk:
```suggestion
postinclude-hooks += $(bootsplash_region_hook)

```

Makefile
```diff
@@ -337,6 +337,29 @@ ifeq ($(FAILBUILD),1)
 $(error cannot continue build)
 endif
 
+# Add bootsplash to BOOTSPLASH/logo.bmp
+define bootsplash_region_hook
```
SergiiDmytruk: Why make it a hook?  `.config` is sourced in line 184 of this file.

SergiiDmytruk (commented):
> Not quite, I think what's key is that there's a black "empty" default image built into the payload, and then there's the customizable BOOTSPLASH region that's set to the Protectli logo. The customer wanted to be able to remove the logo to display the fallback black screen instead.

Got it, thanks.

Makefile
filipleple: [Done](https://github.com/Dasharo/coreboot/pull/771/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R343)

Makefile
filipleple: [Removed](https://github.com/Dasharo/coreboot/pull/771/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R354)

Makefile
filipleple: No longer using a hook

Makefile
filipleple: [Unhooked](https://github.com/Dasharo/coreboot/pull/771/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R340)

mkopec (approved):

PR #781, opened 2025-09-24
edk2/Makefile: Add -v for compiler errors in the output

philipanda:
I find it no fun to try and fix the ways of my coreboot+EDK2 inexperienced self without any errors being printed from EDK2 during the build.

### DIFF
--- a/payloads/external/edk2/Makefile
+++ b/payloads/external/edk2/Makefile
@@ -43,6 +43,9 @@ endif
 
 BUILD_STR += -D BUILD_ARCH=X64
 
+# Enable compiler errors printing from EDK2
+BUILD_STR += -v
+
 #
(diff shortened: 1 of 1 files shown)


miczyg1:
@philipanda you don't need to do that. Simply invoke `make V=1` when building coreboot, and edk2 will not be compiled in quiet mode. Then you will see all compiler errors, if any.

philipanda:
> @philipanda you don't need to do that. Simply invoke `make V=1` when building coreboot, and edk2 will not be compiled in quiet mode. Then you will see all compiler errors, if any.

Thanks, I've checked that `make help` does indeed list that possibility in one of the first lines of the output.
In that case I think exposing this option to `build.sh` would be better suited for the discussion in https://github.com/Dasharo/coreboot/pull/776

PR #784, opened 2025-10-01
.github/workflows/build.yml: fix build_novacustom dependency

filipleple:
Fixes the dependency on deprecated build_novacustom job, replacing it with the new build_novacustom_laptop


philipanda (approved):
NUCBOX builds fine now. The deploy jobs are not being activated, but there is no unknown symbols now, so they too are fine,

PR #786, opened 2025-10-02
mb/novacustom/mtl-h/variants/dgpu/overridetree.cb: Fix dGPU not comin…

mkopec:
…g up after suspend

Fix the root port flags by bringing back the values they were at in release v0.9.1 where suspend worked. This fixes issue where the dGPU would not come back after suspend [1]

[1]: https://github.com/Dasharo/dasharo-issues/issues/1593

*ref: ncm-2010*

### DIFF
--- a/src/mainboard/novacustom/mtl-h/variants/dgpu/overridetree.cb
+++ b/src/mainboard/novacustom/mtl-h/variants/dgpu/overridetree.cb
@@ -17,7 +17,7 @@ chip soc/intel/meteorlake
 			register "pcie_rp[PCH_RP(12)]" = "{
 				.clk_src = 6,
 				.clk_req = 6,
-				.flags = PCIE_RP_LTR | PCIE_RP_HOTPLUG,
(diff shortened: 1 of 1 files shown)


mkopec:
TEST= Boot to Ubuntu 24.04, suspend, and wake. The dGPU comes back up, and the laptop resumes successfully without issue.

mkopec:
Fixes https://github.com/Dasharo/coreboot/pull/786

SergiiDmytruk:
So the use of `PCIE_RP_HOTPLUG` was an unrelated change in b9a010346aa410bcf72c9eaddeb9b9e3f6b07b8e?  Just checking that dGPU will still be available after reboots.

mkopec:
It was an attempt to get the GPU detected more reliably after a reboot but it seems that it's not necessary in the end. The GPU is detected after reboots.

SergiiDmytruk (approved):

PR #790, opened 2025-10-16
configs/config.novacustom_v5*: unset boot resolution, set setup menu …

mkopec:
…resolution

This change makes it so that the boot logo is shown in native resolution and correct aspect ratio, while the setup menu is scaled to 1920x1080 which is bigger and sligtly vertically stretched.

Change-Id: I897bd00fb01414b5ea7e4c3668e54f0eb4d8813d


macpijan:
You have mentioned some issue here regarding setup menu text, maybe we can post picture to demonstrate ?

mkopec:
Sure. Setup menu has 1920x1080 resolution, making it perfectly readable:

image

but the Press F2 for setup menu prompt is at 2880x1800, at which resolution it is a bit difficult to read:

image

macpijan:
> but the Press F2 for setup menu prompt is at 2880x1800, at which resolution it is a bit difficult to read:

Is it different in comparison to what we have had so far, or the same (is it a regression)?

macpijan:
@wessel-novacustom any feedback on these pictures?

wessel-novacustom:
> @wessel-novacustom any feedback on these pictures?

We are totally OK with this. Very good improvement.

macpijan (approved):

PR #791, opened 2025-10-16
mb/{novacustom,clevo}: Bump SMMSTORE size to 512K

mkopec:
Because UEFI DBX keeps growing in size, and we have a lot of UEFI
variables, we've run into an issue where there's not enough free space
to update the DBX. Increase the SMMSTORE region to 512K to mitigate this
issue.

*ref: ncm-2040*


src/mainboard/clevo/adl-p/vboot-rwa.fmd
```diff
@@ -6,12 +6,12 @@ FLASH 32M {
...
 	UNUSED 11M
 
-	RW_MISC 424K {
+	RW_MISC 680K {
 		UNIFIED_MRC_CACHE(PRESERVE) 128K {
 			RECOVERY_MRC_CACHE 64K
 			RW_MRC_CACHE 64K
 		}
-		SMMSTORE(PRESERVE) 256K
+		SMMSTORE(PRESERVE) 512K
```
macpijan, line 14: 1. Is it going to be required for other devices as well?
2. Why it has become a problem now? Anything changed, or simply not spotted before?
3. How much free space we have now after this change?

macpijan:
@mkopec care to explain **why** in the commit description?

src/mainboard/clevo/adl-p/vboot-rwa.fmd
mkopec: > Is it going to be required for other devices as well?

Yes, all of them.

>  Why it has become a problem now?

Because DBX keeps balooning in size.

mkopec:
@SergiiDmytruk I'm having some trouble with capsule update variable migration not working correctly when updating from an image with 256K to an image with 512K SMMSTORE size. Any idea what might be going wrong?

src/mainboard/clevo/adl-p/vboot-rwa.fmd
macpijan: Well, it will be by design, right? I wonder when we would need another size increase.

SergiiDmytruk:
> @SergiiDmytruk I'm having some trouble with capsule update variable migration not working correctly when updating from an image with 256K to an image with 512K SMMSTORE size. Any idea what might be going wrong?

[This](https://github.com/Dasharo/edk2/blob/3afc86b599c663c641d9e8d94be9d47cee675c5b/DasharoPayloadPkg/Library/EfiVarsLib/EfiVarsLib.c#L100-L102) is likely the cause.  It's actually `BlockCount / 2 - 1` (`Fv.Length / BlockSize / 2 - 1`).  The same mistake is in https://github.com/Dasharo/coreboot/blob/4eed42d318e3ce580b2763242d0a4f12f36ce840/util/smmstoretool/fv.c#L78-L80
so you can check using `smmstoretool`.

src/mainboard/clevo/adl-p/vboot-rwa.fmd
SergiiDmytruk: Growing SMMSTORE from 4 to 8 blocks (64 KiB each) increases variable storage part of it from 1 to 3 blocks, that is from 64 KiB to 192 KiB.  Size of DBX is 23.44 KiB, so those extra 128 KiB should provide plenty of room.

In case 128 KiB of flash make a difference for some platforms, SMMSTORE size can also be 384 KiB in size, with variable store itself being 2 blocks or 128 KiB.

mkopec:
Hmm seems to be failing even with this change:

```
Found SMM Store information
block size: 0x10000
number of blocks: 0x8
communication buffer: 0x49804000
communication buffer size: 0x10000
MMIO address of store: 0xFF020000
Created SMMSTORE info hob
ValidateFvHeader: Variable Store Length does not match
BootMode: Boot with default settings
...
RecoverVariableStorage: Recovery is not needed or not possible.
NvStorageBase:0xFF020000, NvStorageSize:0x80000
FvbInitialize: The FVB Header is not valid.
FvbInitialize: Installing a correct one for this volume.
InstallProtocolInterface: D1A86E3F-0707-4C35-83CD-DC2C29C891A3 0
InstallProtocolInterface: 09576E91-6D3F-11D2-8E39-00A0C969723B 493DC180
InstallProtocolInterface: 8F644FA9-E850-4DB1-9CE2-0B44698E8DA4 493DC128
SmmStoreInitInstance: Created a new instance
```

SergiiDmytruk:
> Hmm seems to be failing even with this change:

What value did you use?  `Fv.Length / BlockSize / 2 - 1` is the correct number of blocks instead of `1` assumed currently, so [that line](https://github.com/Dasharo/edk2/blob/3afc86b599c663c641d9e8d94be9d47cee675c5b/DasharoPayloadPkg/Library/EfiVarsLib/EfiVarsLib.c#L102) should end up looking like this:
```c
  VarStoreHdr->Size      = BlockSize * (Fv.Length / BlockSize / 2 - 1) - VolHdrLen;
```

mkopec:
Okay your exact copied line worked :sweat_smile: 

```c
  VarStoreHdr->Size      = BlockSize * (Fv.Length / BlockSize / 2 - 1) - VolHdrLen;
```

macpijan (dismissed):
Is it possible to send https://github.com/Dasharo/coreboot/pull/791/commits/05b6f5505ed7213793fa93abd35b1fa452bb6a83 upstream right away?

util/smmstoretool/fv.c
```diff
@@ -77,7 +77,7 @@ bool fv_init(struct mem_range_t fv)
 		.Signature = EfiAuthenticatedVariableGuid,
 		// Actual size of the storage is block size, the rest is
 		// Fault Tolerant Write (FTW) space and the FTW spare space.
```
SergiiDmytruk:
```suggestion
		// Actual size of the storage is `block_count / 2 - 1`, the rest is
		// Fault Tolerant Write (FTW) space (1 block) and the FTW spare space (half of all blocks).
```

mkopec:
Huh, looks like upstream already fixed it: https://review.coreboot.org/c/coreboot/+/88427/5/util/smmstoretool/fv.c

macpijan:
@mkopec 

> Huh, looks like upstream already fixed it: https://review.coreboot.org/c/coreboot/+/88427/5/util/smmstoretool/fv.c

Is it possible that we then backport this specific commit, instead of merging an alternative solution? If we add some more value to the upstream one, we should backport, create our value-added on top of that, and send upstream as well.

mkopec:
sure @macpijan , removed our changes and backported the commit from upstream.

macpijan (approved):

SergiiDmytruk (approved):

mkopec:
TEST=Update via UEFI Capsule Update from the previous RC and check that EFI settings are preserved correctly. Try setting a variable with smmstoretool and verify that the setting is correctly read by firmware.

PR #797, opened 2025-10-23
superio/ite/it8659e/superio.c: Force HWM/PECI logic reset on warm boot

filipleple:
PECI temperature reporting on this chip was stuck at a fixed value after reboots, making the fan speed tied to it stay the same regardless of load. It worked as expected after a power-off.

Add an explicit reset of the HWM/PECI logic to prevent the chip from staying in an undefined state after platform reboots.

*ref: prot-1953*

### DIFF
--- a/src/superio/ite/it8659e/superio.c
+++ b/src/superio/ite/it8659e/superio.c
@@ -2,6 +2,7 @@
 
 #include <device/device.h>
 #include <device/pnp.h>
+#include <delay.h>
 #include <pc80/keyboard.h>
 #include <superio/conf_mode.h>
 #include <superio/ite/common/env_ctrl.h>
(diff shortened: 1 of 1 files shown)


filipleple:
Looks like the CI fail has to do with the test setup rather than this change

macpijan:
> Looks like the CI fail has to do with the test setup rather than this change

True. Did it work in previous PRs lately? I just clicked to rerun, let's see.

src/superio/ite/it8659e/superio.c
```diff
@@ -22,6 +23,13 @@ static void it8659e_init(struct device *dev)
 		res = probe_resource(dev, PNP_IDX_IO0);
 		if (!conf || !res)
 			break;
+
+		/* Force HWM/PECI logic reset on warm boot */
```
SergiiDmytruk, line 27: This doesn't actually check for warm boot, so maybe the comment should say something like
```suggestion
		/* Always force HWM/PECI logic reset to avoid it getting stuck on after a warm boot */
```

Also, is the reset bit defined in `IT8659_H_V0.7.2_industrial_20190917.pdf` or that's a wrong datasheet?

filipleple:
Well, this is fairly awkward. I saw it reproduce on the v0.9.2 binary and immediately got caught up in testing different writes, but didn't notice that it just works on the new `dasharo` branch. When I finally found a write that didn't break (do) anything, I assumed that it was the solution...

PR #798, opened 2025-10-24
configs/config.novacustom_*: enable debug SMI

mkopec:
Enable debug SMI, which works around issue
https://github.com/Dasharo/dasharo-issues/issues/1664.

This is a workaround and should be removed once a proper solution to the issue is found.

TEST=Enable SMM_BWP, reboot and try toggling and saving settings 10 times, verify no form submission errors occur.

Change-Id: I47ce515248c3c3339a4e91f183e32ab79ae2bdcc

### DIFF
--- a/src/mainboard/clevo/adl-p/Kconfig
+++ b/src/mainboard/clevo/adl-p/Kconfig
@@ -124,6 +124,10 @@ config SOC_INTEL_CSE_SEND_EOP_EARLY
 config VBOOT_FWID_VERSION
 	default "\$(CONFIG_LOCALVERSION)"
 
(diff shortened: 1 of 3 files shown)


SergiiDmytruk (commented):
Would adding
```kconfig
# A terrible workaround for SMI issues when BWP is on until a better solution is found.
config DEBUG_SMI
	default y if BOOTMEDIA_SMM_BWP
```
to `src/mainboard/novacustom/Kconfig` work the same?  I would put it there instead of defconfigs where it looks like a leftover from debugging and has no explanation.

mkopec:
Done @SergiiDmytruk . And I do agree it's quite terrible :|

macpijan (approved):

PR #801, opened 2025-10-28
CBnT without TXT

mkopec:
Make ACM support independent from TXT being enabled.

Depending on customer requirements, CBnT may need to be enabled without
also enabling TXT. In this configuration, only the Boot Guard component
of CBnT is actively used.

TEST=Boot V540TU without TXT, without and with ACMs, confirm platform boots and Boot Guard works if ACMs are present. Confirm https://github.com/Dasharo/dasharo-issues/issues/1674 does not occur. Confirm TXT is disabled.


mkopec:
Fixes https://github.com/Dasharo/dasharo-issues/issues/1674 , because HVCI is seemingly incompatible with TXT. I've confirmed that this PR fixes the issue and that TXT is disabled.

SergiiDmytruk:
I wonder what's the state of bits 1 (`Enable VMX in SMX operation.`) and 2 (`Enable VMX outside SMX operation.`) of `IA32_FEATURE_CONTROL` MSR when Hyper-V doesn't like SMX being enabled.  Maybe the error message isn't precise and the real problem is that SMX is only partially enabled.

Also, can BootGuard be affecting this?

The changes look good, I'm just trying to better understand why they are necessary.

miczyg1:
> I'm just trying to better understand why they are necessary.

When the TXT support in FSP is buggy, or you don't want to have TXT, it is better to have it disabled, rather than make coreboot attempt to initialize TXT and risk boot loops if any error occurs.

Documentation/security/intel/txt.md
```diff
@@ -92,9 +92,9 @@ correct state. If it's not the SINIT ACM will reset the platform.
 ### Configuring Intel TXT in Kconfig
 Enable ``INTEL_TXT`` and set the following:
 
-``INTEL_TXT_BIOSACM_FILE`` to the path of the BIOS ACM provided by Intel
+``INTEL_BIOSACM_FILE`` to the path of the BIOS ACM provided by Intel
```
miczyg1: Before CBnT there were 2 separate BIOS ACMs:

- Boot Guard S-ACM 
- TXT BIOS ACM

I'm not sure renaming helps here... Especially that this ACM is specifically used for TXT in coreboot.

src/soc/intel/cannonlake/Kconfig
```diff
@@ -331,13 +331,13 @@ config SOC_INTEL_COMMON_OC_WDT_TIMEOUT_SECONDS
 
 if INTEL_TXT
```
miczyg1: Some of the settings moved under `INTEL_ACM` umbrella, so it should be `if INTEL_ACM`. Needs fixing probably across all socs.

Documentation/security/intel/txt.md
mkopec: Ok, renamed back: https://github.com/Dasharo/coreboot/pull/801/commits/3c0763e2b043507bc753f635b5c91269721ef9c8

src/soc/intel/cannonlake/Kconfig
mkopec: Fiuxed in https://github.com/Dasharo/coreboot/pull/801/commits/2c92c08ad167b6fd89a8e5764b173d7b679b3b5a . Only that one config option needed to be moved under `if INTEL_ACM`

mkopec:
@miczyg1 could you take a look again?

Documentation/security/intel/txt.md
macpijan: `INTEL_BIOSACM_FILE` is no longer there, so I assume this resolves it.

src/soc/intel/cannonlake/Kconfig
macpijan: INTEL_TXT was present in the MTL and CNL Kconfigs only. This should be fine.

macpijan (approved):

PR #810, opened 2025-11-12
msi: use `intel-microcode` instead of `dasharo-blobs`

Tonux599:
Some boards (like the MSI boards) have not been getting microcode from the `intel-microcode` sub-module, instead being hard coded to `dasharo-blobs`, which is out of date.

The PR reverts the commit pinning it to `dasharo-blobs`, as well as bumping the `intel-microcode` sub-module to it's latest release.


miczyg1:
The whole mess was caused by bad integration.

Like I instructed here: https://github.com/Dasharo/dasharo-issues/issues/1079#issuecomment-2396518431
It was supposed to be changed in board config, not the sources that generate microcode from tree...

miczyg1 (approved):

miczyg1:
NUC BOX/laptop builds probably not going to work when PRs are opened from a fork due to missing secret token...

PR #818, opened 2025-12-09
Btg capsule check

mkopec:
Expose HFSTS registers via SMBIOS to allow EDK2 to determine if platform is fused or not.

Bump EDK2 for Boot Guard signing key check: https://github.com/Dasharo/edk2/pull/285

*ref: ncm-2095*


SergiiDmytruk (approved):

SergiiDmytruk:
The force-push was for rebasing onto latest `dasharo`.

PR #821, opened 2025-12-12
Use all active TPM banks for Measured Boot

SergiiDmytruk:
issue: https://github.com/Dasharo/dasharo-issues/issues/982
*ref: prot-1808*


src/security/vboot/tpm_common.c
```diff
@@ -47,19 +48,37 @@ tpm_result_t vboot_extend_pcr(struct vb2_context *ctx, int pcr,
 	enum vb2_hash_algorithm algo = tlcl_get_family() == TPM_1 ?
 		VB2_HASH_SHA1 : VB2_HASH_SHA256;
 
+	int i, j;
+	struct tpm_digest digests[ENABLED_TPM_ALGS_NUM + 1];
+	for (i = 0, j = 0; i < ENABLED_TPM_ALGS_NUM; ++i) {
```
filipleple, line 50: @SergiiDmytruk is it intended that `j` is not incremented in this loop?

src/security/vboot/tpm_common.c
```diff
@@ -47,19 +48,37 @@ tpm_result_t vboot_extend_pcr(struct vb2_context *ctx, int pcr,
...
+			printk(BIOS_WARNING,
+			       "vboot: Not extending %s digest (buffer in %s is too small).\n",
+			       vb2_get_hash_algorithm_name(alg), __func__);
+			continue;
+		}
+
+		/* Extending the same data to all banks.  It either gets truncated, fits
+		   perfectly or is padded with zeroes. */
+		digests[j].hash = buffer;
+		digests[j].hash_type = algo;
```
filipleple: We do want to use `algo`, defined [before](https://github.com/Dasharo/coreboot/pull/821/changes#diff-50f6d62bb2106e7c65381f5d4f6fa87699065c636d827c69e0e63739e6eb2b23R48) this loop, not `alg` defined and used inside this loop?

src/security/tpm/tspi/tspi.c
```diff
@@ -269,40 +270,84 @@ tpm_result_t tpm_measure_region(const struct region_device *rdev, uint8_t pcr,
...
+				printk(BIOS_ERR, "TPM: Error extending hash.\n");
+				return TPM_CB_HASH_ERROR;
+			}
+		}
+		if (vb2_digest_finalize(&ctx, digest, digest_len)) {
+			printk(BIOS_ERR, "TPM: Error finalizing hash.\n");
 			return TPM_CB_HASH_ERROR;
 		}
+
+		digests[j].hash = digest;
```
filipleple: Admittedly, \*_according to LLM_\*: 

The function computes multiple digests by iterating algorithms, but it finalizes each digest into the same stack buffer `uint8_t digest[TPM_PCR_MAX_LEN]` and then stores `digests[j].hash = digest`. All entries end up pointing to the same memory and therefore all banks will extend with the last computed digest, with truncation effects for smaller algorithms. This breaks both PCR values and event log correctness when more than one bank is active.

I'm not sure if this is the case, feel free to disregard

filipleple (changes requested):

src/security/vboot/tpm_common.c
SergiiDmytruk: Oh, of course not, thanks for catching this.  Turns out `CONFIG_VBOOT_NO_TPM` is set for some Protectli boards, so vboot doesn't extend any PCRs making this harder to notice in manual tests.

https://github.com/Dasharo/coreboot/compare/5c23cb0c62e5820a865d6f710d9c1515e2edf5a2..69aee15ab38c6f4dec05d62b97e059bd0441cef0#diff-50f6d62bb2106e7c65381f5d4f6fa87699065c636d827c69e0e63739e6eb2b23R66

src/security/vboot/tpm_common.c
SergiiDmytruk: Must be `alg`, thanks.  Missing increment of `j` hid this mistake.

https://github.com/Dasharo/coreboot/compare/5c23cb0c62e5820a865d6f710d9c1515e2edf5a2..69aee15ab38c6f4dec05d62b97e059bd0441cef0#diff-50f6d62bb2106e7c65381f5d4f6fa87699065c636d827c69e0e63739e6eb2b23R65

src/security/tpm/tspi/tspi.c
SergiiDmytruk: LLM is actually right in this case.  This affects only FMAP measurement, which I did notice (it's the first measurement) but incorrectly attributed this weirdness to vboot extending the same value regardless of the digest type and didn't check the code.

https://github.com/Dasharo/coreboot/compare/5c23cb0c62e5820a865d6f710d9c1515e2edf5a2..69aee15ab38c6f4dec05d62b97e059bd0441cef0#diff-154fbdfb0b02f8a77b8f64a02671aaf370e8b5a17383722e9e852e78ccf3d9bd

SergiiDmytruk (commented):
Updated the code and reran tests on VP4650.  Also checked that build with `CONFIG_VBOOT_NO_TPM=n` produces correct PCR values.

mkopec:
@filipleple could you take a look if your questions are addressed?

filipleple (approved):

filipleple:
LGTM, will merge when the CI goes through after rebasing.

EDIT: merged via CLI, somehow github didn't pick up on it

PR #823, opened 2026-01-07
Support loading and processing on-disk capsules during the same boot

SergiiDmytruk:
This makes it possible to disable ME via HMRFPO before applying an update, as on-disk capsules don't rely on preserving RAM's state across reboots.

As usual, better review commits in order.

This also enables publication of SMBIOS table necessary to determine whether a platform is fused on Hardkernel Odroid, where the changes were tested but hit an error because capsules thought the device has been fused.

First half of https://github.com/Dasharo/dasharo-issues/issues/1438#issuecomment-3703953941 could be used for context.

EDK PR: https://github.com/Dasharo/edk2/pull/287
issue: https://github.com/Dasharo/dasharo-issues/issues/1438
*ref: dsh-1130*


mkopec:
Code looks valid. I haven't tested it yet.

mkopec (approved):

mkopec:
Tested, left comment in the edk2 PR: https://github.com/Dasharo/edk2/pull/287#issuecomment-3727580265

PR #828, opened 2026-01-20
mb/protectli/vault_adl_n: vp2440: reenable ASPM and disable clk gating

mkopec:
It was found that this change improves idle power draw by 1.8 W, while also ensuring that the platform behaves the same as VP2430 with regards to ASPM.

Upstream-Status: Pending
Change-Id: Icd174e4694e86bc98d4d9b68e1cc124bc4b22d2f

### DIFF
--- a/src/mainboard/protectli/vault_adl_n/mainboard.c
+++ b/src/mainboard/protectli/vault_adl_n/mainboard.c
@@ -85,6 +85,15 @@ void mainboard_silicon_init_params(FSP_S_CONFIG *params)
 		 * params->PcieRpEnableCpm[10] = 1;
 		 */
 		params->PcieRpEnableCpm[11] = 1;
+
+		/*
+		 * VP2430 does not have clock gating, because not all CLKREQs are
+		 * connected. In tests, this meant that network performance was higher.
+		 * Disable clock gating manually on VP2440 to make it perform in line
+		 * with VP2430.
+		 */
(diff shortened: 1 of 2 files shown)


mkopec:
See the comment [here](https://github.com/Dasharo/coreboot/blob/e4f7fa45c49af24aaec3772d3f10dfea927c93d3/src/soc/intel/alderlake/fsp_params.c#L1053-L1064) for more context on the relation between CLKREQ and clock gating. 

It was observed that if ASPM is enabled, NIC performance was higher on VP2430 than on VP2440. VP2430 doesn't connect CLKREQ signals to all PCIe peripherals, so clock gating was being disabled. It was determined that ASPM + clock gating enabled at the same time were the cause of reduced performance of VP2440.

This change switches from one to another to improve power draw and bring VP2440 closer to VP2430 wrt performance and power draw.

mkopec:
Measurements done on the AC power adapter input:

- latest dasharo branch (clock gating, no ASPM): 8.9 watts
- dasharo + disabled clock gating + enabled ASPM: 7.1 watts

That's at idle, no disks, netbooted into DTS, with one ethernet cable plugged in. About 20% idle power draw reduction.

pietrushnic:
@mkopec this is very good information. Congratulations.

filipleple (approved):
LGTM, nice improvement!

PR #835, opened 2026-01-28
drivers/efi/capsules.c: auto-set attempt_slot_b on capsule detection

filipleple:
*ref: prot-2021*

### DIFF
--- a/src/drivers/efi/capsules.c
+++ b/src/drivers/efi/capsules.c
@@ -23,6 +23,9 @@
 #include <Guid/FmpCapsule.h>
 #include <IndustryStandard/WindowsUxCapsule.h>
 
+/* Weak default - platforms can override to prepare redundant slots */
+__weak void efi_capsule_update_prepare_redundant_slot(void) { }
+
 /*
  * Overview
  *
@@ -801,7 +804,7 @@ BOOT_STATE_INIT_ENTRY(BS_DEV_INIT, BS_ON_EXIT, parse_capsules, NULL);
 
 #endif
 
-static void enable_capsule_smi(void *unused)
+static void enable_capsule_update_path(void *unused)
 {
 	uint32_t ret;
 
@@ -824,6 +827,10 @@ static void enable_capsule_smi(void *unused)
 
 	printk(BIOS_INFO, "%sabled capsule update SMI handler\n",
 	       ret == SMMSTORE_RET_SUCCESS ? "En" : "Dis");
+
+	/* For redundancy-enabled platforms, ensure booting from Slot B */
+	if (full_flash_access)
+		efi_capsule_update_prepare_redundant_slot();
 }
 
-BOOT_STATE_INIT_ENTRY(BS_POST_DEVICE, BS_ON_ENTRY, enable_capsule_smi, NULL);
+BOOT_STATE_INIT_ENTRY(BS_POST_DEVICE, BS_ON_ENTRY, enable_capsule_update_path, NULL);
--- a/src/drivers/efi/capsules.h
+++ b/src/drivers/efi/capsules.h
@@ -24,4 +24,12 @@ static inline void efi_add_capsules_to_bootmem(void) { }
 
 #endif
 
+/*
+ * Called when UEFI update capsules are detected.
+ *
+ * Platform-specific code can override this to prepare for capsule updates,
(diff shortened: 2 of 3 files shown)


src/drivers/efi/capsules.c
```diff
@@ -659,6 +662,9 @@ static void coalesce_capsules(struct block_descr block_chain, uint8_t *target)
 	}
 
 	printk(BIOS_INFO, "capsules: found %d capsule(s).\n", uefi_capsule_count);
+
+	if (uefi_capsule_count > 0)
+		efi_capsule_update_prepare_redundant_slot();
```
SergiiDmytruk: `coalesce_capsules()` is not the right place, better rename `enable_capsule_smi()` at the bottom and do it there using `full_flash_access` for a condition (to take on-disk capsules into account).

src/soc/intel/common/block/rtc/rtc.c
```diff
@@ -43,6 +44,20 @@ void rtc_conf_set_bios_interface_lockdown(void)
 					PCR_RTC_CONF_BILD);
 }
 
+#if CONFIG(INTEL_TOP_SWAP_OPTION_CONTROL)
+#include <drivers/efi/capsules.h>
```
SergiiDmytruk: Including this at the top should do no harm.  `if (CONFIG(INTEL_TOP_SWAP_OPTION_CONTROL))` can also be done inside the function.

SergiiDmytruk (changes requested):
Will do for now, but coreboot assuming a successful update can lead to unexpected user-visible behaviour.  I think we need an issue to [do it in EDK](https://github.com/Dasharo/edk2/pull/286#issuecomment-3786053739).

filipleple:
added the request for future improvement: https://github.com/Dasharo/dasharo-issues/issues/1756

src/drivers/efi/capsules.h
```diff
@@ -12,6 +12,14 @@ void efi_parse_capsules(uintptr_t *base, size_t *size);
...
 void efi_add_capsules_to_bootmem(void);
 
+/*
+ * Called when UEFI update capsules are detected.
+ *
+ * Platform-specific code can override this to prepare for capsule updates,
+ * e.g., to switch to a redundant firmware slot for safe updates.
+ */
+void efi_capsule_update_prepare_redundant_slot(void);
+
```
SergiiDmytruk: CI is failing with:
```
src/soc/intel/common/block/rtc/rtc.c:52:6: error: no previous prototype for 'efi_capsule_update_prepare_redundant_slot' [-Werror=missing-prototypes]
   52 | void efi_capsule_update_prepare_redundant_slot(void)
```
Maybe that's why `#if` was used in `rtc.c`.  I think moving this declaration out of `#if` should be fine.  At most one such function should be defined outside of `capsules.c` anyway.

SergiiDmytruk (approved):
Force-pushed to rebase and remove a stray empty line.

I also figured out why last time testing redundancy changes confused me, there is console output despite:
```
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_8 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_7 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_6 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_5 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_4 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_3 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_2 is not set
# CONFIG_DEFAULT_CONSOLE_LOGLEVEL_1 is not set
CONFIG_DEFAULT_CONSOLE_LOGLEVEL_0=y
CONFIG_DEFAULT_CONSOLE_LOGLEVEL=0
```
That must be because of `debug_level` in CMOS options and I took output as a sign that wrong firmware is being executed.

filipleple:
@SergiiDmytruk thanks! 

And yes it confused me a couple times as well, IIRC that's why we moved from CMOS backend to NVRAM storage of options, with the latter you can be sure what exact config state the platform will be after flashing, with the former it's not so obvious

PR #829, opened 2026-01-21
Add fwupd support / LVFS script

filipleple:
Integrate the functionality of capsule_cabinet.sh into capsule.sh. Cabinets can now be created using `./capsule.sh create_cabinet capsule.cap`. Prior functionality and calls of capsule.sh remain unaffected.

Upstream-Status: Inappropriate [Dasharo downstream]

issue: https://github.com/Dasharo/dasharo-issues/issues/1432
_ref: DSH-1123_

PR #848, opened 2026-02-18
Fwupd lvfs script

filipleple:
Continuing from: https://github.com/Dasharo/coreboot/pull/829

issue: https://github.com/Dasharo/dasharo-issues/issues/1432
_ref: DSH-1123_


capsule.sh
```diff
@@ -294,6 +316,94 @@ EOF
...
+
+    local date vendor version
+    date=$(stat -c %w "$capsule" 2>/dev/null | cut -d ' ' -f 1)
+    if [ -z "$date" ] || [ "$date" = "-" ]; then
+        date=$(stat -c %y "$capsule" 2>/dev/null | cut -d ' ' -f 1)
+    fi
+
+    local id=com.${vendor}.${CONFIG_MAINBOARD_SMBIOS_PRODUCT_NAME}.${CONFIG_MAINBOARD_VERSION}.system.firmware
+    id=${id// /_}
+    id=${id////_}
```
SergiiDmytruk: This needs to be moved down to run after `$vendor` is set.

SergiiDmytruk (changes requested):
One block should be moved and probably need to switch target branch.

capsule.sh
filipleple: Righto, [moved](https://github.com/Dasharo/coreboot/pull/848/changes/b36e25a8d506d383002dbd357b4647c9c7f38486) and rebased onto `capsule-v2`

SergiiDmytruk (approved):

PR #851, opened 2026-02-20
Redundancy: Switch to Slot B upon successful update

### LINKED ISSUE #1756: Redundancy: Ensure that after a capsule update, the new firmware is booted
### The problem you're addressing (if any)

Currently, automation _is being/has been_ added in coreboot to force booting from the update partition/Slot B. It's hooked at the point where coreboot probes whether a capsule is present: https://github.com/Dasharo/coreboot/pull/835.

This is OK for a start, but not ideal since EDK can reject capsule processing for various reasons. The next, more robust iteration would be implementing the mechanism in EDK2.

### Describe the solution you'd like

Looking at nvramtool, setting the flag from EDK would take finding `cmos.layout` in CBFS and parsing it to find where the flag and checksum are in CMOS, then updating both, which doesn't seem that complicated.

### Where is the value to a user, and who might that user be?

The value would be that the flag isn't unnecessarily flipped in the scenario where the capsule is later rejected by EDK2, and firmware isn't updated.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

mkopec:
See: https://github.com/Dasharo/edk2/pull/305

Fixes: https://github.com/Dasharo/dasharo-issues/issues/1756

*ref: prot-2030*


SergiiDmytruk (approved):
Force-pushed to bump EDK revision.

PR #852, opened 2026-02-24
NovaCustom ADL-P: TPM init and Cbnt fixes after FSP change

filipleple:
After switching to IoT FSP, TPM 2.0 would no longer initialize. To fix this, GPIO configuration had to be moved earlier, before ramstage enumeration, mimicking Protectli VP66XX.

Upstream-Status: Pending


mkopec (dismissed):

mkopec (dismissed):

src/security/intel/acm/Kconfig
```diff
@@ -48,4 +48,34 @@ config INTEL_TXT_SINIT_SIZE
...
+	  then removes it from the final coreboot.rom.
+
+	  This serves two purposes:
+	  - The build fails at cbfstool time if there is insufficient CBFS space
+	    for the ACM, rather than silently producing an image that cannot be
+	    provisioned.
+	  - After the stub is removed, the provisioning tool can insert the real
+	    ACM into the freed, correctly-aligned slot.
+
+	  The stub is NOT a functional ACM and must not be left in the image.
```
mkopec, line 69: Does this mean someone building their own firmware from our defconfigs needs to remove it?

Is it really bad if it's left in there? I thought TxT / BtG would just not work. or does it bootloop?

filipleple: @mkopec it's taken care of by the buildsystem automatically, the placeholders are removed in the end, leaving only ACM-sized "holes". Probably should've worded it more clearly

I'm not 100% sure, but the way I remember things we did initially try to use placeholders, but they caused some trouble so we reverted to not adding ACMs at all within coreboot, only with the provisioning script.

But that might be wrong/confusing with a different mechanism.

mkopec: ah it removes it. sorry, missed that part.

mkopec (approved):

PR #861, opened 2026-03-11
Windows update

mkopec:
*ref: dsh-1124*


capsule_cabinet.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
-# A script to generate cabinets for fwupd / LVFS
+# Generate a firmware update cabinet compatible with both LVFS/fwupd and Windows Update.
```
SergiiDmytruk, line 2: Just as a note in case you've missed it: `capsule_cabinet.sh` is now a part of `capsule.sh` (https://github.com/Dasharo/coreboot/commit/213216e16d9037b8f606d27a3955d99e76f4184e) on `capsules-v2` branch.

PR #862, opened 2026-03-15
build: allow overriding docker UID for rootless environments

Suraj-kumar00:
### Description
This PR enables building the coreboot firmware using a rootless Docker setup, fixing [Dasharo/dasharo-issues#1197](https://github.com/Dasharo/dasharo-issues/issues/1197).

Currently, `build.sh` uses `-u $UID` for the `docker run` commands. In a rootless Docker setup, mapping the host's UID into the container causes user-namespace conflict errors. Because `$UID` is a read-only variable in Bash, it cannot be overridden safely from the command line.

This introduces a `DOCKER_UID` environment variable that falls back to `$UID`. Developers using rootless setups can now reproduce builds by running:

```bash
DOCKER_UID=root ./build.sh qemu
```

Because of rootless Docker user namespaces, the container will run as `root` internally and bypass the permission constraint. The final `.rom` output file will still be safely owned by the host's standard user.

### Testing

- **Standard Docker (Regression Test):**  
  ```bash
  BUILD_TIMELESS=1 EC_BOARD_VENDOR=clevo EC_BOARD_MODEL=nv40mz ./build.sh
  ```
  Builds successfully.

- **Rootless Docker:**  
  ```bash
  BUILD_TIMELESS=1 DOCKER_UID=root EC_BOARD_VENDOR=clevo EC_BOARD_MODEL=nv40mz ./build.sh
  ```
  Successfully completes the build.

- **Verification:**  
  The SHA256 hashes of the resulting `ec.rom` files match perfectly between both build environments, confirming reproducible builds without permission errors.


philipanda:
Looks like you've copied most of the PR comment from the EC comment. The Test does not really make sense here.


I've tested that myself on a rootless docker installation.
The build does succeed without issues when we set the DOCKER_UID on your revision.
The SHA is not the same though.
The `sha256sum` I get on your change (e2e8d5a2ee60):
  - `5c577cd52dc17c571ac498d51d5cfccd65a1e3130b1164830d28187e09960087`.
  
The hash on the commit prior (5db5cf9d6ba3) (built on a rootful docker installation): 
  - `a8bb3f34875f8c564dbabb8b805d97062c7ddbef63384a6e9159a37891458636`

We can check what has changed in the binary using [romscope](https://github.com/Dasharo/romscope).
The reports I got show quite a lot of changes different than revision or timestamps.
Could you help with interpreting them @mkopec?
[report.zip](https://github.com/user-attachments/files/26208454/report.zip)
You can try to run the same checks yourself too @Suraj-kumar00 

### Test procedure

1. On a fresh Ubuntu 22.04 machine, as a non-root user Install rootless Docker
  ```
  curl -fsSL https://get.docker.com/rootless | sh
  export PATH=$HOME/bin:$PATH
  export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock
  ```
  
2. Verify it's truly rootless
  ```
  docker info | grep -i rootless
  ```
  
3. Setup repo
  ```
  git clone https://github.com/dasharo/coreboot.git && cd coreboot
  git remote add suraj https://github.com/Suraj-kumar00/coreboot.git
  ```
4. Test if it fails on rootless
  ```
  ./build.sh qemu
  ```
5. Test if fix works
  ```
  git checkout suraj fix/rootless-docker-uid
  ./build.sh qemu
  ```
6. Build without fix and compare SHA256
  ```
  sha256sum qemu_q35_v0.2.1-rc1.rom
  git checkout suraj fix/rootless-docker-uid~1
  ./build.sh qemu
  sha256sum qemu_q35_v0.2.1-rc1.rom
  ```

PR #873, opened 2026-04-15
mainboard/dell: align shared product string to 7010/9010

WuXieSec:
## Summary
Follow-up low-risk alignment for #1182: update shared product string for the currently combined 7010/9010 SFF board target.

## Change
- `src/mainboard/dell/snb_ivb_workstations/Kconfig`
  - `default "OptiPlex 9010" if BOARD_DELL_OPTIPLEX_9010`
  - -> `default "OptiPlex 7010/9010" if BOARD_DELL_OPTIPLEX_9010`

## Why
Current board target is used for shared 7010/9010 SFF support. This keeps user-visible naming aligned with support scope while bigger separation work continues.

Related issue: https://github.com/Dasharo/dasharo-issues/issues/1182


src/mainboard/dell/snb_ivb_workstations/Kconfig
```diff
@@ -34,7 +34,7 @@ config MAINBOARD_DIR
 	default "dell/snb_ivb_workstations"
 
 config MAINBOARD_PART_NUMBER
-	default "OptiPlex 9010" if BOARD_DELL_OPTIPLEX_9010
+	default "OptiPlex 7010/9010" if BOARD_DELL_OPTIPLEX_9010
```
miczyg1: This will break the identification of the boards. Even if the codebase is shared, the builds should be separate so the string `"OptiPlex 9010"` and `"OptiPlex 7010"` still matches the vendor BIOS and our detection in the Dasharo Tools Suite.

```suggestion
	default "OptiPlex 7010" if BOARD_DELL_OPTIPLEX_7010
	default "OptiPlex 9010" if BOARD_DELL_OPTIPLEX_9010
```

A new Kconfig option BOARD_DELL_OPTIPLEX_7010 should be introduced, even if it just selects different MAINBOARD_PART_NUMBER only. Also, an appropriate entry in Kconfig.name file must be added.

miczyg1 (changes requested):

WuXieSec:
Thanks, you were right. I reworked this to keep separate 7010 and 9010 identification strings, introduced `BOARD_DELL_OPTIPLEX_7010`, and added the matching `Kconfig.name` entry. For now both options still point at the existing shared variant directory, so the change stays minimal while preserving the expected product string for DTS detection.

Updated in `98ffd527e3`.

WuXieSec:
@miczyg1 已按您的要求完成修改（commit 98ffd527e3）：引入了 BOARD_DELL_OPTIPLEX_7010 选项，在 Kconfig.name 添加了对应条目，保持了分开的产品字符串。麻烦重新审查，谢谢。

PR #875, opened 2026-04-16
Capsules V2 release changes

SergiiDmytruk:
These are RC1s for MSI boards, new CI checks of defconfigs and build system updates to support provisioning of capsule root keys.

*ref: dsh-1176*


.github/scripts/capsule-versioning-check.sh
```diff
@@ -0,0 +1,89 @@
...
+# rare and obvious to the user
+mapfile -t configs < <(grep -l 'EFI_UPDATE_CAPSULE' configs/config.*)
+for config in "${configs[@]}"; do
+    new_config=$(git show "$new_tag:$config")
+    old_config=$(git show "$old_tag:$config")
+
+    # check that `CONFIG_EDK2_CAPSULES_V2=y` is not added to
+    # `CONFIG_DRIVERS_EFI_UPDATE_CAPSULES=y` without
+    # `CONFIG_EDK2_CAPSULES_V2_TRANSITION=y`
+    if uses_v1_capsules "$old_config" &&
```
philipanda, line 83: Is `CONFIG_DRIVERS_EFI_UPDATE_CAPSULES` supposed to be disabled on full, non-transition V2 releases?
The checks don't allow `CONFIG_DRIVERS_EFI_UPDATE_CAPSULES=y` && `CONFIG_EDK2_CAPSULES_V2=y` after the transition is finished.

.github/scripts/capsule-versioning-check.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# check all configs to not figure out mapping tag name to a config name, some
+# false-positives are possible due to overlapping releases, but that should be
+# rare and obvious to the user
+mapfile -t configs < <(grep -l 'EFI_UPDATE_CAPSULE' configs/config.*)
+for config in "${configs[@]}"; do
+    new_config=$(git show "$new_tag:$config")
+    old_config=$(git show "$old_tag:$config")
+
+    # check that `CONFIG_EDK2_CAPSULES_V2=y` is not added to
```
philipanda, line 80: Could also check if the transition is not used without `CONFIG_EDK2_CAPSULES_V2`.

.github/scripts/capsule-versioning-check.sh
```diff
@@ -0,0 +1,89 @@
...
+function transitions_to_v2() {
+    local config=$1
+    echo "$config" | grep -qFx "CONFIG_EDK2_CAPSULES_V2_TRANSITION=y"
+}
+
+new_tag=$1
+
+old_tag=$(get_previous_release "$new_tag")
+if [ -z "$old_tag" ]; then
+    echo "warning: no previous release for '$tag', skipping checks."
```
filipleple: I don't know if `$tag` is defined in this context? I think it's only defined within `get_previous_release()` [here](https://github.com/Dasharo/coreboot/pull/875/changes#diff-a4082e2c4c2e597ca86a95505489d4954d3186de19ad83c83827a22fbad1cfe3R10)

.github/workflows/checks.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   check-upstream-status:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
```
filipleple, line 10: LLM input - I don't know gh actions to the point where I can tell if it's valid, please feel free to resolve if this is nonsense:

```
  Risk - shallow clone breaks git show

  tag-check.yml uses actions/checkout@v4 (shallow, depth=1 by default). The subsequent git fetch --tags --force --prune fetches tag
  refs but not necessarily the commit trees they point to. git show "$old_tag:$config" will fail with fatal: Not a valid object name on
   a shallow clone. The checkout step needs fetch-depth: 0:

  - uses: actions/checkout@v4
    with:
      fetch-depth: 0
```

filipleple (changes requested):

.github/scripts/capsule-versioning-check.sh
```diff
@@ -0,0 +1,89 @@
...
+
+fail=0
+
+# check all configs to not figure out mapping tag name to a config name, some
+# false-positives are possible due to overlapping releases, but that should be
+# rare and obvious to the user
+mapfile -t configs < <(grep -l 'EFI_UPDATE_CAPSULE' configs/config.*)
+for config in "${configs[@]}"; do
+    new_config=$(git show "$new_tag:$config")
+    old_config=$(git show "$old_tag:$config")
```
filipleple, line 78: This fails fails if file didn't exist in old_tag, so with set -eu a newly-added board config might always fail CI at tag time?

.github/workflows/checks.yml
philipanda: IMO doesn't make sense as this job is just the upstream-status check. The `capsule-versioning-check.sh` is not being used here.

.github/workflows/tag-check.yml
philipanda, line 16: @filipleple regarding https://github.com/Dasharo/coreboot/pull/875/changes/BASE..9e0383799a1501f667e63a7a8904de2d5ad3ac39#r3099406485, here the tags are being fetched, so that works

.github/scripts/capsule-versioning-check.sh
philipanda: I think there is no way a platforms config file would not exist on its release tag. If there is a tag for a platform where there is no config file for it, I think panicking is expected

filipleple: right, forgot that in this context they both have to be tags for this specific platform. the old_tag won't be some random older tag from before the platform was added.

.github/scripts/capsule-versioning-check.sh
philipanda: should probably be `new_tag`

SergiiDmytruk: Yes, I renamed variables for consistency at some point and introduced this mistake.

https://github.com/Dasharo/coreboot/compare/9e0383799a1501f667e63a7a8904de2d5ad3ac39..e72877ebc25579a978d268ab52b829349c211a0b#diff-a4082e2c4c2e597ca86a95505489d4954d3186de19ad83c83827a22fbad1cfe3R65

.github/scripts/capsule-versioning-check.sh
SergiiDmytruk: That can't happen because https://github.com/Dasharo/coreboot/blob/a155c21d7a7d3f397acbbd6f661c0d188eeee92c/payloads/external/edk2/Kconfig.dasharo#L442-L445
and this script probably shouldn't validate state of `Kconfig` files.

.github/scripts/capsule-versioning-check.sh
SergiiDmytruk: Not sure which combination of options you think is problematic.  `CONFIG_DRIVERS_EFI_UPDATE_CAPSULES=y` will always be there when capsules are enabled.  `uses_v1_capsules()`  returns non-zero for a config with V2 capsules, so this condition will not be entered.

SergiiDmytruk (commented):
https://github.com/Dasharo/coreboot/compare/9e0383799a1501f667e63a7a8904de2d5ad3ac39..e72877ebc25579a978d268ab52b829349c211a0b adds:
 * check that `CONFIG_EDK2_CAPSULES_V2_TRANSITION=y` is not added to a V2 config
 * check that `CONFIG_EDK2_CAPSULES_V2=y` is not removed
 * usage
 * redirection or errors and warnings to stderr

SergiiDmytruk:
Added `./capsule.sh resign` which we'll need for MSI release.

filipleple (approved):

SergiiDmytruk:
Added 159ead466c16c6250f21050817e9272fa0ad02c1 to enable the following usage scenario:
 1. `./capsule.sh box`.
 2. Put production capsule signing keys to `gencap/keys/`.
 3. Use `capsule.sh` + `gencap/` to resign capsules outside of build tree.

philipanda (approved):

PR #894, opened 2026-05-19
mb/protectli/vault_ehl: Enable RST1 button signal GPIO input

philipanda:
The button sits on a multi-function pin of the IT8613E. Enabled the chip's GPIO block in devicetree.cb, muxed the pin to GP12 (away from PCIRST1#), and turned on its internal pull-up to make the signal go back to 1 after the button is released.

### DIFF
--- a/src/mainboard/protectli/vault_ehl/bootblock.c
+++ b/src/mainboard/protectli/vault_ehl/bootblock.c
@@ -4,16 +4,32 @@
 #include <bootblock_common.h>
 #include <superio/ite/common/ite.h>
 #include <superio/ite/it8613e/it8613e.h>
+#include <device/pnp_ops.h>
+#include <superio/ite/common/ite_gpio.h>
+
 
 #define GPIO_DEV PNP_DEV(0x2e, IT8613E_GPIO)
 #define UART_DEV PNP_DEV(0x2e, IT8613E_SP1)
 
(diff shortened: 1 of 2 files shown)


philipanda:
Test: Release, Press, Release, Press:
```console
root@DasharoToolsSuite:~# dd if=/dev/port bs=1 skip=$((0xa00)) count=1 2>/dev/null | hexdump
0000000 0004                                   
0000001
root@DasharoToolsSuite:~# dd if=/dev/port bs=1 skip=$((0xa00)) count=1 2>/dev/null | hexdump
0000000 0000                                   
0000001
root@DasharoToolsSuite:~# dd if=/dev/port bs=1 skip=$((0xa00)) count=1 2>/dev/null | hexdump
0000000 0004                                   
0000001
root@DasharoToolsSuite:~# dd if=/dev/port bs=1 skip=$((0xa00)) count=1 2>/dev/null | hexdump
0000000 0000                                   
0000001
root@DasharoToolsSuite:~# 
```

Also works on zarhus-openwrt with https://github.com/zarhus/protectli-openwrt-factory-reset-service. The service requires a change to disable the path where it returns 1 for VP2420 as unsupported. Holding down the `RST1` button for ~10s causes a reboot.

philipanda:
@mkopec restored

mkopec (approved):
Looks fine to me.

PR #895, opened 2026-05-19
mb/protectli/vault_ehl: Enable RST1 button signal GPIO input

philipanda:
The button sits on a multi-function pin of the IT8613E. Enabled the chip's GPIO block in devicetree.cb, muxed the pin to GP12 (away from PCIRST1#), and turned on its internal pull-up to make the signal go back to 1 after the button is released.


mkopec (approved):
LGTM

PR #900, opened 2026-05-21
configs/config.protectli_vp2430_no_emmc:  config variant without eMMC

wiktormowinski:
Adds ENABLE_EMMC option (default y) to support VP2430e, a hardware variant without eMMC. SMBIOS product name gets an "e" suffix when the option is disabled.

`ScsEmmcEnabled` is set by `soc_silicon_init_params()` via `is_devfn_enabled(PCH_DEVFN_EMMC)` before `mainboard_silicon_init_params()` is called, so overriding it directly in FSP UPD at the end of `mainboard_silicon_init_params()` is sufficient.

`pcidev_path_on_root(PCH_DEVFN_EMMC)->enabled` is also cleared to prevent the device from appearing in ACPI tables.


philipanda (approved):

philipanda:
Seems the tag `protectli_vp2430_v0.9.1-rc1` was not updated after the force pushes

philipanda (approved):
tag fixed for rc2

PR #916, opened 2026-06-14
Drtm payload

hanetzer:
Not ready for merge. Just getting it out there for discussion reasons.


SergiiDmytruk:
You actually want to rebase onto `dasharo-25.12` (shouldn't be much effort).  I think `dasharo` is based on coreboot 24.12.  `dasharo-25.12` is also closer to upstream which should get the patches later.

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,120 @@
...
+#include <slrt/slrt.h>
+
+#include <lib.h>
+
+static uintptr_t payload_start, payload_size;
+
+/* For SELF, prog->start and prog->size are not set so obtain it differently. */
+void platform_segment_loaded(uintptr_t start, size_t size, int flags)
+{
+	/* FIXME: need to differentiate between payload and other loaded segments */
```
SergiiDmytruk, line 18: Add another flag like `SEG_ELF` in `rmodule_load()`?  Or maybe can make `prog->start` and `prog->size` be set and get rid of this function?

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,120 @@
...
+#include <stdlib.h>
+#include <security/tpm/tis.h>
+#include <cpu/amd/msr.h>
+#include <slrt/slrt.h>
+
+#include <lib.h>
+
+static uintptr_t payload_start, payload_size;
+
+/* For SELF, prog->start and prog->size are not set so obtain it differently. */
```
SergiiDmytruk, line 15:
```suggestion
/* For CBFS_TYPE_SELF (Simple ELF), prog->start and prog->size are not set so obtain it differently. */
```

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,120 @@
...
+	struct slr_table *slrt;
+	struct slr_entry_dl_info *dl_info;
+	struct slr_entry_amd_info *amd_info;
+	struct slr_entry_hdr *end;
+
+	/*
+	 * Check if we're on 32b platform.
+	 * TODO: add support for 64b?
+	 */
+	assert(sizeof(skl) == 4);
```
SergiiDmytruk, line 58: Since it's not needed yet, can replace this with `_Static_assert()` if the code makes assumptions about 32 bits.  Otherwise, can be removed, `Kconfig` constraints should be enough.

src/soc/amd/common/skl/Kconfig
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+config LAUNCH_DRTM_PAYLOAD
+	bool "Launch DRTM payload before the real one"
+	default n
+	depends on MEMORY_MAPPED_TPM
```
SergiiDmytruk, line 6: Should also depend on `ARCH_RAMSTAGE_X86_32`.

The implementation is only for AMD at this point, but not sure how to check for it reliably.

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,120 @@
...
+	slrt->size = 16;
+	slrt->max_size = 0;
+
+	dl_info = (struct slr_entry_dl_info *)(skl + bootloader_data_offset + 16); // TODO: Do Smarter
+	dl_info->hdr.tag = SLR_ENTRY_DL_INFO;
+	dl_info->hdr.size = sizeof(struct slr_entry_dl_info);
+	dl_info->dlme_base = payload_start;
+	dl_info->dlme_size = payload_size;
+	dl_info->dlme_entry = (uint32_t)prog->entry;
+	dl_info->bl_context.bootloader = SLR_BOOTLOADER_GRUB; // TODO: BOOTLOADER_COREBOOT?
```
SergiiDmytruk, line 90: I suppose worth adding an ID to avoid funny-looking code, although I don't remember it being utilized beyond simple checks.

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,120 @@
...
+
+	memset(slrt, 0, 64*KiB - (skl - (void *)slrt));
+
+	slrt->magic = 0x4452544d;
+	slrt->revision = 0x1;
+	slrt->architecture = 0x2;
+	slrt->size = 16;
+	slrt->max_size = 0;
+
+	dl_info = (struct slr_entry_dl_info *)(skl + bootloader_data_offset + 16); // TODO: Do Smarter
```
SergiiDmytruk, line 84:
```suggestion
	dl_info = (struct slr_entry_dl_info *)slrt->entries;
```

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,120 @@
...
+	amd_info->hdr.tag = SLR_ENTRY_AMD_INFO;
+	amd_info->hdr.size = sizeof(*amd_info);
+	slrt->size += amd_info->hdr.size;
+
+	end = next_tag(amd_info);
+	end->tag = SLR_ENTRY_END;
+	end->size = sizeof(struct slr_entry_hdr);
+	slrt->size += end->size;
+
+	/* TODO: DRTM TPM event log and SKL hash(es) */
```
SergiiDmytruk, line 103: This is about adding `SLR_ENTRY_LOG_INFO` and `SLR_ENTRY_ENTRY_POLICY`.  GRUB's sources in https://github.com/TrenchBoot/grub/tree/tb-dev/grub-core/loader/slaunch can be used as an example for such things.

src/soc/amd/common/skl/Kconfig
hanetzer: As it currently stands, because its in the amd/common dir, it seems its only reachable if SOC_AMD_COMMON is set. That being said, would you consider this location to be fine for now?

src/soc/amd/common/skl/skinit.c
hanetzer: I'm not sure how much this actually does for us. we're including the binary file, not an elf, payload type raw.

src/soc/amd/common/skl/Kconfig
SergiiDmytruk: Yes, we can find a better place later and update commits as necessary.

src/soc/amd/common/skl/skinit.c
SergiiDmytruk: SKL blob isn't a payload, EDK is coreboot's payload and simultaneously a DLME for DRTM.  `build/UEFIPAYLOAD.fd` isn't actually of ELF type but that's the type it somehow gets in CBFS:
```
FMAP REGION: COREBOOT
Name                           Offset     Type           Size   Comp
cbfs_master_header             0x0        cbfs header        32 none
fallback/payload               0x80       simple elf    1377908 none
...
```

src/soc/amd/common/skl/skinit.c
hanetzer: some questions about that. most of this seems to be coming from a slparams struct, like the tpm event log. this seems to be at an arbitrary memory location, is that correct? Also, some of the data that's measured seem to be coming directly from grub's linux cmdline, which doesn't exist at this point in bootup with the coreboot->drtm flow, should I just avoid these entries for now?

src/soc/amd/common/skl/skinit.c
hanetzer: Derp. yeah I should have known that, but I've not touched c in any meaningful way since the x370 port.

src/soc/amd/common/skl/skinit.c
hanetzer: pushed this suggestion back into the commit the comment existed in, no need for churn I think.

src/soc/amd/common/skl/skinit.c
hanetzer: I just mean we're literally taking the binary skl.bin and shoving it in as a raw file, the elf is never touched aside from being used to produce the bin.

src/soc/amd/common/skl/skinit.c
hanetzer: Done

src/soc/amd/common/skl/skinit.c
SergiiDmytruk: SKL is a different kind of payload and that's probably what confuses you.  In TCG's terminology it's called DCE (DRTM Configuration Environment), maybe this abbreviation should be used for the file name.

`platform_segment_loaded()` is not invoked in response to `cbfs_load()`, it's invoked earlier when `fallback/payload` is being loaded before `platform_prog_run()` is called.  The control-flow is defined by these functions of the state machine: https://github.com/Dasharo/coreboot/blob/1617d9eaacc02944bdd27310a8d39ff84e81701e/src/lib/hardwaremain.c#L192-L208

`payload_load()` prepares `global_payload` and `payload_run()` indirectly invokes `platform_prog_run()` in this file with a pointer to `global_payload`, see: https://github.com/Dasharo/coreboot/blob/1617d9eaacc02944bdd27310a8d39ff84e81701e/src/lib/prog_loaders.c#L150-L208

src/soc/amd/common/skl/skinit.c
SergiiDmytruk: > most of this seems to be coming from a slparams struct, like the tpm event log.

Yes, slparams struct is a global state which gets updated and accessed from different places.  It can be filled differently depending on how things are loaded (i.e., MultiBoot, Linux boot or EFI boot and whether it's Intel or AMD).

> this seems to be at an arbitrary memory location, is that correct?

I think so.  Allocating memory using cbmem should work and will take care of reserving the memory from the use by EDK or OS.

> Also, some of the data that's measured seem to be coming directly from grub's linux cmdline, which doesn't exist at this point in bootup with the coreboot->drtm flow, should I just avoid these entries for now?

The set of things to measure depends on what gets started and what data it consumes, so Linux params are not in scope when DRTM is done in firmware.  In case of EDK, that's things like ACPI, SMBIOS and at least coreboot table in CBMEM.  Some of CBMEM will change each boot and can't be added, pointers changing values in otherwise stable data can also be a problem, can try with a few first (`cbmem_entry_find(CBMEM_ID_CBTABLE)` and the same for `CBMEM_ID_ACPI` and `CBMEM_ID_SMBIOS`).  Will likely need a list of CBMEM entries that are stable between boots, but the exact set is yet to be determined, so only need generic implementation for now.

PR #926, opened 2026-06-19
Drtm payload 25.12

### LINKED ISSUE #68: coreboot implementation for AMD
This task will implement support for performing DRTM between coreboot and its payload, in this case edk2. While the implementation of dynamic launch will be in coreboot, the payload must be able to handle unusual state after the hand-off. It also has to carefully manage the access to memory, as well as control when and how other cores are allowed to be started. Finally, the payload binary must contain information that would tell coreboot that it can be started through dynamic launch.

The implementation may reuse parts of [PoC](https://github.com/pcengines/coreboot/compare/63fc56b90d50f62d57b3627b98876975bab7cd6e...pcengines:coreboot:drtm_payload).

- AMD only, but `cpu/amd/pi` isn't proper place for this code
- Many commits were added just for edk2 payload (at least all with SMM in the name), those should be present in Dasharo already so they don't have to be ported
- SKL revision wasn't specified in the config, default branch was always used. This must be changed to have reproducible builds
- SLRT needs to be added, old code uses SKL tags
- PoC TODOs: include headers directly from SKL, 64b (optional?), reserve space for TPM event log
- Documentation in coreboot: description of new option, use cases, supported/tested families and payload requirements (common and AMD-specific)
- If Dasharo gains support for newer AMD platforms: changes required for PSP involvement in SKINIT - discovery, implementation and documentation
- Send patches to upstream coreboot

Form: PR to https://github.com/Dasharo/coreboot, and if needed also https://github.com/TrenchBoot/secure-kernel-loader
Proof: logs from debug version of SKL

hanetzer:
Replacement of #916, can't change my branch, just the target. Hoping to avoid losing the comments there. Sorry for the noise.


src/acpi/dsdt_top.asl
```diff
@@ -7,7 +7,7 @@
 #include <vendorcode/google/chromeos/acpi/chromeos.asl>
 #endif
 
-External (\_SB.NAPE, MethodObj)
+External (\_SB.PCI0.NAPE, MethodObj)
```
SergiiDmytruk: Can't simply rename this, the change is [in upstream](https://review.coreboot.org/c/coreboot/+/89479) and will break AMD Turin (see `soc/amd/turin_poc/acpi/ioapic_routing.asl`).

Getting rid of `PCI0` will affect other methods, so maybe permit both names and pick one via `CondRefOf`.

src/acpi/dsdt_top.asl
```diff
@@ -30,9 +30,10 @@ Method (_PIC, 1)
 {
 	/* Remember the OS' IRQ routing choice. */
 	PICM = Arg0
-	If (CondRefOf (\_SB.NAPE))
+
+	If (CondRefOf (\_SB.PCI0.NAPE))
 	{
-		\_SB.NAPE()
+		\_SB.PCI0.NAPE(Arg0 & 1)
```
SergiiDmytruk: For compatibility, the parameter should be dropped in favour of using `PICM`.

src/include/acpi/acpi_ivrs.h
```diff
@@ -170,6 +170,11 @@
 #define CAP_OFFSET_10_MSI_NUM_PPR_SHIFT		27
 #define CAP_OFFSET_10_MSI_NUM_PPR		(0x1f << CAP_OFFSET_10_MSI_NUM_PPR_SHIFT)
 
+#define PCI_DEVID(BUS, DEV, FN) ( \
```
SergiiDmytruk: `src/include/device/pci_def.h` seems like a better place for this macro, that's where `PCI_DEVFN` is defined.

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,101 @@
...
+	cbfs_load(CONFIG_CBFS_PREFIX "/drtm_payload", skl, 64*KiB);
+
+	bootloader_data_offset = ((struct sl_header *)skl)->bootloader_data_offset;
+	slrt = (struct slr_table *)(skl + bootloader_data_offset);
+
+	memset(slrt, 0, 64*KiB - (skl - (void *)slrt));
+
+	slrt->magic = 0x4452544d;
+	slrt->revision = 0x1;
+	slrt->architecture = 0x2;
```
SergiiDmytruk: Better to use constants like in https://github.com/TrenchBoot/grub/blob/tb-dev/include/grub/slr_table.h.

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,101 @@
...
+	slrt->size = 16;
+	slrt->max_size = 0;
```
SergiiDmytruk: `slrt->size` is `sizeof(*slrt)` and `slrt->max_size` should be `64*KiB - bootloader_data_offset` (`64*KiB - (skl - (void *)slrt)` above is a more complicated way of computing the same).

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,101 @@
...
+	struct slr_entry_amd_info *amd_info;
+	struct slr_entry_hdr *end;
+
+	/*
+	 * Check if we're on 32b platform.
+	 * TODO: add support for 64b?
+	 */
+	_Static_assert(sizeof(skl) == 4);
+
+	hexdump(prog, sizeof(*prog));
```
SergiiDmytruk: Don't see a need to keep this.

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,101 @@
...
+	/*
+	 * APs have to be in wait-for-SIPI state for at least 1000 cycles before
+	 * SKINIT. Send INIT now and assume that loading SKL from CBFS is long
+	 * enough.
+	 */
+	lapic_send_ipi_others(LAPIC_INT_LEVELTRIG | LAPIC_INT_ASSERT | LAPIC_MT_INIT);
+
+	skl = memalign(64*KiB, 64*KiB);
+
+	cbfs_load(CONFIG_CBFS_PREFIX "/drtm_payload", skl, 64*KiB);
```
SergiiDmytruk: Both functions can fail.  This function should `die()` in such cases, because simply returning will apparently result in repeated invocation with possibly weird side-effects.

.gitmodules
```diff
@@ -85,3 +85,6 @@
 [submodule "3rdparty/dasharo-blobs"]
 	path = 3rdparty/dasharo-blobs
 	url = https://github.com/Dasharo/dasharo-blobs.git
+[submodule "3rdparty/secure-kernel-loader"]
+	path = 3rdparty/secure-kernel-loader
+	url = https://github.com/hanetzer/trenchboot-secure-kernel-loader
```
SergiiDmytruk: If we don't need to have SKL as a submodule, I think it's worth avoiding adding one.  Adding submodules results in them being checked out for no use and also require setting up a mirror in upstream coreboot.

SergiiDmytruk (commented):
Reviewed the changes in more detail.

.gitmodules
hanetzer: yeah I can probably drop this and some related bits, this was when I was trying for explicitly using slrt.h from skl itself; when its done via payloads/external it doesn't exist early enough for compiling the ramstage.

src/acpi/dsdt_top.asl
hanetzer: I pulled it from apu2_fix_iommu and rebased it as-is, in fact everything before the drtm stuff is from that branch. if its not needed at all I could just drop those commits.

src/soc/amd/common/skl/skinit.c
hanetzer: Yeah definitely, was mostly just 'banging it out' to get it to actually start talking to me.

src/soc/amd/common/skl/skinit.c
hanetzer: the assert or the hexdump?

src/include/acpi/acpi_ivrs.h
hanetzer: honestly unsure if I even need the stuff before the drtm commits at this point. I should give that a test, try to focus effort in one area.

src/acpi/dsdt_top.asl
hanetzer: see above

src/soc/amd/common/skl/skinit.c
SergiiDmytruk: The `hexdump()` invocation.  The lines above it are shown by GitHub as a context.

src/acpi/dsdt_top.asl
hanetzer: ugh. guess I do need these; rebased to dasharo-25.12 and dropped them, no longer 'functions' as it does. I'm *not* an acpi person at all, I'll ask the original author about it.

hanetzer:
I think one problem I'm really having is mapping secure-launch terminology to coreboot parts. This is outside my normal area of 'doing coreboot' stuff, I'd much rather and be more in my comfort zone porting new platforms but I'll take what I can get. If you could explain, for example, what part of this stack maps to DLME, for example, I'd probably have an easier time of it moving forward.

SergiiDmytruk:
> I think one problem I'm really having is mapping secure-launch terminology to coreboot parts. This is outside my normal area of 'doing coreboot' stuff, I'd much rather and be more in my comfort zone porting new platforms but I'll take what I can get. If you could explain, for example, what part of this stack maps to DLME, for example, I'd probably have an easier time of it moving forward.

Just saw this comment and I think you figured it out by now, but anyway:
 - DCE preamble is coreboot
 - DL Event is `SKINIT` instruction or rather its handler in CPU/PSP
 - DCE is SKL
 - DLME is EDK2 and by extension whatever it runs (which is the purpose of putting DRTM in the firmware)

src/soc/amd/common/skl/skinit.c
hanetzer: did. rearranged it a little to avoid extra maths.

hanetzer:
> Just saw this comment and I think you figured it out by now, but anyway:
> 
> * DCE preamble is coreboot
> * DL Event is `SKINIT` instruction or rather its handler in CPU/PSP
> * DCE is SKL
> * DLME is EDK2 and by extension whatever it runs (which is the purpose of putting DRTM in the firmware)

Thank you. I'm honestly crap with terminology at times so having a definitive 'this is that' list helps. I guess if edk2 loads grub, grub also goes onto the dlme listing? and also later the kernel?

SergiiDmytruk:
> I guess if edk2 loads grub, grub also goes onto the dlme listing? and also later the kernel?

Yes, although it makes sense to start Linux directly by the firmware in this scenario to execute less code in the early boot environment (Secure Boot could validate GRUB and it has lockdown mode, but there is still a lot of code in there).

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,117 @@
...
+	amd_info = next_entry(dl_info);
+	amd_info->hdr.tag = SLR_ENTRY_AMD_INFO;
+	amd_info->hdr.size = sizeof(*amd_info);
+	slrt->size += amd_info->hdr.size;
+
+	log_info = next_entry(amd_info);
+	log_info->hdr.tag = SLR_ENTRY_LOG_INFO;
+	log_info->hdr.size = sizeof(*log_info);
+	slrt->size += log_info->hdr.size;
+	ce = cbmem_entry_find(CBMEM_ID_TPM2_TCG_LOG);
```
SergiiDmytruk, line 87: This event log is for SRTM.  DRTM's measurements form an independent chain that can also be reset if DRTM is re-started, so the log needs to be separate (`cbmem_add()`).

src/soc/amd/common/skl/skinit.c
```diff
@@ -0,0 +1,117 @@
...
+	if (ce) {
+		log_info->addr = (uintptr_t)cbmem_entry_start(ce);
+		log_info->size = cbmem_entry_size(ce);
+		log_info->format = SLR_DRTM_TPM20_LOG; // TODO: Support 1.2?
```
SergiiDmytruk, line 91: > // TODO: Support 1.2?

coreboot, SKL and EDK support TPM1.2 and the cost for it here is a bit of initialization (see it [in GRUB](https://github.com/TrenchBoot/grub/blob/b5517787efc2359908c268e7bad5eb0bd1043525/grub-core/loader/slaunch/txt.c#L538-L562); despite the name that function is used on AMD as well), so may as well do it.  TPM2.0 log just needs to be zeroed, its size in GRUB is 32 KiB regardless of TPM family.

src/soc/amd/common/skl/skinit.c
hanetzer: hmm. Would I need to define a new CBMEM_ID then? CBMEM_ID_SRTM_LOG or so?

SergiiDmytruk: > hmm. Would I need to define a new CBMEM_ID then?

Yes, I think `cbmem_add()` requires it.  Updating https://github.com/Dasharo/coreboot/blob/dasharo/src/commonlib/bsd/include/commonlib/bsd/cbmem_id.h is two places should be enough.

> CBMEM_ID_SRTM_LOG or so?

"DRTM", maybe `CBMEM_ID_TPM_DRTM_LOG` to be less ambiguous.

PR #931, opened 2026-06-26
payloads/edk2/Kconfig.dasharo: bump EDK for EC serial flushing fix

philipanda:
https://github.com/Dasharo/edk2/pull/314

### DIFF
--- a/payloads/external/edk2/Kconfig.dasharo
+++ b/payloads/external/edk2/Kconfig.dasharo
@@ -4,7 +4,7 @@ config EDK2_REPOSITORY
 	default "https://github.com/Dasharo/edk2"
 
 config EDK2_TAG_OR_REV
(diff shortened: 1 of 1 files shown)


filipleple (approved):
can be merged as well once CI/commit message is amended

philipanda:
@filipleple the commit in question is not a part of this PR :/
The commit is also tagged as novacustom tgl v1.6.0-rc3, so Im not sure we can amend it now?

filipleple:
@philipanda 
> the commit in question is not a part of this PR :/
The commit is also tagged as novacustom tgl v1.6.0-rc3, so Im not sure we can amend it now?

image

Welp, I guess that's not in the scope of this PR so it can be merged, we can think about it when integrating the `tgl_rel` branch with `dasharo` at some point

PR #932, opened 2026-06-29
config.protectli_vp2420*: remove rc suffix

filipleple:


filipleple:
seems to be wrong base, vp2420 is pre-rebase

PR #936, opened 2026-07-03
configs/config.novacustom_v5*: Bump version to next rc1

mkopec:
Change-Id: I48d1801361efbefc05cff3be1210e24bbcebb248


filipleple (approved):
CI is complaining, but that's about the gigabyte board so out of scope

PR #941, opened 2026-07-15
Novacustom tgl v1.6.0 SBOM fixes

philipanda:
Merge after https://github.com/Dasharo/coreboot/pull/937

### DIFF
--- a/src/sbom/Makefile.mk
+++ b/src/sbom/Makefile.mk
@@ -84,6 +84,16 @@ endif
 # binary files in order to give more complete/detailed information inside the SBOM file.
 # These files are either in src/sbom/ or build/sbom (if they are generated).
 swid-files-$(CONFIG_SBOM_ME) += $(if $(CONFIG_SBOM_ME_GENERATE), $(build-dir)/intel-me.json, $(CONFIG_SBOM_ME_PATH))
+# ME/TXE on IFWI platforms (Apollo Lake/Gemini Lake): SBOM_ME cannot be
+# enabled there through Kconfig, because it depends on HAVE_ME_BIN and the
+# TXE is stitched inside the IFWI image instead of being a standalone
+# binary. Include the generated ME/TXE tag whenever an SBOM is built on
+# such platforms (same approach as the edk2 payload override above).
+ifeq ($(CONFIG_NEED_IFWI),y)
+ifneq ($(CONFIG_SBOM_ME),y)
+swid-files-y += $(build-dir)/intel-me.json
+endif
+endif
 swid-files-$(CONFIG_SBOM_PAYLOAD) += $(if $(CONFIG_SBOM_PAYLOAD_GENERATE),$(payload-swid),$(if $(CONFIG_PAYLOAD_EDK2),$(payload-swid),$(CONFIG_SBOM_PAYLOAD_PATH)))
 # TODO think about just using one CoSWID tag for all intel-microcode instead of one for each. maybe put each microcode into files entity of CoSWID tag?
 swid-files-$(CONFIG_SBOM_INTEL_MICROCODE) += $(patsubst 3rdparty/intel-microcode/intel-ucode/%, $(build-dir)/intel-microcode-%.json, $(filter 3rdparty/intel-microcode/intel-ucode/%, $(cpu_microcode_bins)))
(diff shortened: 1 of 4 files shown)


src/sbom/Makefile.mk
```diff
@@ -170,26 +170,41 @@ $(build-dir)/compiler-%.json: $(src-dir)/compiler-%.json | $(build-dir)/goswid
...
 		| grep -m1 -Eo 'ME[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+' \
-		| sed 's/^ME//'); \
-	if [ -n "$$me_ver" ]; then \
-		sed -i "s/<software_version>/$$me_ver/" $@; \
-	else \
-		sed -i "/software-version/d" $@; \
-	fi
+		| sed 's/^ME//');
+	if [ -z "$$me_ver" ]; then
+		off=$$(grep -aboF -m1 '$$MN2' "$$me" | head -1 | cut -d: -f1); \
```
miczyg1: @philipanda 

$MN2 is a magic string for a component manifest. Manifests may precede other components than ME, thus the version at offset 8 from that magic string may not always indicate the version of ME, but a different component. Also, there are multiple ocurrences of $MN2, what if the first $MN2 is not from ME?

We would need to extract the exact partition and subpartition with ME. There is cse_serger utility in cbfstool, which could be used to extract the correct partition and then extract the version from manifest.

```
$ ./build/cse_serger 3rdparty/dasharo-blobs/novacustom/ns5x_tgl/me.bin  print-layout -v 1.7
 * CSE Layout

ROM Bypass: 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 
Size: 0x40
Redundancy: 0x0
Checksum: 0x21bd59c6
Data partition offset: 0x1f0000
Data partition size: 0x87000
BP1 offset: 0x2000
BP1 size: 0x1ee000
BP2 offset: 0x0
BP2 size: 0x0
BP3 offset: 0x277000
BP3 size: 0x1ee000
BP4 offset: 0x0
BP4 size: 0x0
BP5 offset: 0x0
BP5 size: 0x0
Temp base addr: 0x0
Temp base size: 0x0
FLOG offset: 0xffffffff
FLOG size: 0xffffffff

$ dd if=3rdparty/dasharo-blobs/novacustom/ns5x_tgl/me.bin of=cse_main.bin bs=$((0x1ee000)) skip=$((0x277000)) count=1 iflag=skip_bytes 2> /dev/null

$ ./build/cse_serger cse_main.bin dump -n NFTP
Dumping NFTP in NFTP

$ grep -aboF -m1 '$MN2' NFTP | head -1 | cut -d: -f1
384

$ dd if=NFTP skip=392 bs=1 count=8|od -A n -t u2 |xargs|tr \  .
15.0.47.2473
```

1.  `./build/cse_serger 3rdparty/dasharo-blobs/novacustom/ns5x_tgl/me.bin  print-layout -v 1.7`

Print the layout of the ME partitions. Depending on the SOC, the partition layout version may be 1.6 or 1.7. IIRC 1.6 is pre-TGL, TGL and newer has 1.7. For ME 11 or earlier, the process would be different too and would require a different utility and commands.

The CSE Main program is placed at the last BPDT partiion, so we have to extract it. Command below

2.  `dd if=3rdparty/dasharo-blobs/novacustom/ns5x_tgl/me.bin of=cse_main.bin bs=$((0x1ee000)) skip=$((0x277000)) count=1 iflag=skip_bytes 2> /dev/null` Extract the BPDT with CSE Main program.

3. `./build/cse_serger cse_main.bin dump -n NFTP` extracts the CSE Main program subpartition (NFTP) from given BPDT.

4. Now we can search `NFTP` file extracted in previous step to extract version from NFTP manifest which is at the beginning of the NFTP:

offset = `grep -aboF -m1 '$MN2' NFTP | head -1 | cut -d: -f1`

`dd if=NFTP skip=$((offset + 8)) bs=1 count=8|od -A n -t u2 |xargs|tr \  .`

Note the `tr` command replaces space character `\ ` with a `.` to give `15.0.47.2473`. Of coruse you may continue with the hexdump commands you currently have, once you know the offset to $MN2

philipanda: ~~What if we would extract all using all `$MN2` occurences and only get the one that matches the csme versioning scheme (a.b.c.d)? Do you think it would catch anything else than ME?
In the TGL ME I'm using here to test the commands only one occurence matches.~~

*edit that doesn't make sense, we're reading a binary file at that point

philipanda: @miczyg1 I think this should work: https://github.com/Dasharo/coreboot/pull/941/commits/eb105c5378f791b764dd9d35cbcdb06d55cf66b7

It works for me on NV41MZ TGL and extracts the correct CSME version number

miczyg1: > @miczyg1 I think this should work: https://github.com/Dasharo/coreboot/commit/eb105c5378f791b764dd9d35cbcdb06d55cf66b7

Almost. This won't work well for, e.g. Protectli VP46xx, because it uses BPDT format 1.6. We can either:

- use `CONFIG_CSE_BPDT_VERSION` (currently only set for ADL, so would need to be propagated to relevant SoCs)
- use `CONFIG_ME_SPEC` to get proper BPDT, if SOC_INTEL_CSE_HAVE_SPEC_SUPPORT is set

If `CONFIG_ME_SPEC` is going to be used, then:

- if SOC_INTEL_CSE_HAVE_SPEC_SUPPORT not set, then BPDT version is none, because ME images did not have BPDT yet, use cse_fpt utility
- else if CONFIG_ME_SPEC >= 15, then BPDT version is  1.7
- else (every other case, so CONFIG_ME_SPEC > 11 and CONFIG_ME_SPEC < 15), then BPDT version is 1.6
- Apollo Lake and Gemini Lake is a different pair of shoes too, must use ifwitool on the CONFIG_IFWI_FILE_NAME file

If we would ever enable SBOMs for older platforms, it may not work as it currently is done.

Example of dumping NFTP on FW6 ME image (version 11.x):

```
./build/cse_fpt fw6_me.bin dump -n NFTP
```

The rest will be the same, i.e. extraction of version by searching $MN2:

```
$ grep -aboF -m1 '$MN2' NFTP | head -1 | cut -d: -f1
980
$ dd if=NFTP skip=988 bs=1 count=8 2>/dev/null|od -A n -t u2 |xargs|tr \  .
11.8.95.4551
```

Of course none of the above will work if CONFIG_USE_ME_CLEANER is set

Example of dumping NFTP on VP2410 (Gemini Lake - TXE version 4.x):

```
$ ./build/ifwitool 3rdparty/dasharo-blobs/protectli/vault_glk/ifwi.bin extract -f NFTP -n NFTP
Sub-Partition NFTP(7) stored in NFTP.
$ grep -aboF -m1 '$MN2' NFTP | head -1 | cut -d: -f1
1652
$ dd if=NFTP skip=1660 bs=1 count=8 2>/dev/null|od -A n -t u2 |xargs|tr \  .
4.0.50.2083
```

philipanda: > Of course none of the above will work if CONFIG_USE_ME_CLEANER is set

We would use `CONFIG_ME_BIN_PATH` and `CONFIG_IFWI_FILE_NAME` for extracting the ME version. For example for most Dasharo firmware the CONFIG_ME_BIN_PATH would point to `3rdparty/dasharo-blobs`, not to `build/` so ME_CLEANER shouldn't modify these binaries, but rather work on their copies right?
Look at `src/southbridge/intel/common/firmware/Makefile.mk:90-98`, it works on `coreboot.pre` not on `CONFIG_ME_BIN_PATH`.

I've assumed it's the case here: https://github.com/Dasharo/coreboot/pull/941/changes/b792e94ddefa651e9d89c27545944da82c75293e

mkopec (approved):
LGTM, but I'm going to let @miczyg1 resolve his open thread

philipanda:
> LGTM, but I'm going to let @miczyg1 resolve his open thread

Note that we've got 4 stacked branches here, all are targeting the release branch. There is no PR to merge the release branch to `dasharo` yet.
- `dasharo`
- `novacustom_tgl_rel_v1.6.0`
- `novacustom_tgl_v1.6.0_sbom`
- `novacustom_tgl_v1.6.0_sbom_update`
- `novacustom_tgl_v1.6.0_sbom_fixes`

src/sbom/Makefile.mk
miczyg1: Right. I rather meant that if one uses pre-cleaned ME binary. Indeed setting CONFIG_USE_ME_CLEANER alone would not cause problems.

src/sbom/Makefile.mk
```diff
@@ -170,22 +179,83 @@ $(build-dir)/compiler-%.json: $(src-dir)/compiler-%.json | $(build-dir)/goswid
...
+        sbom-me-nftp-tool := $(CSE_FPT)
+    endif
+endif
+
+$(build-dir)/intel-me.json: $(src-dir)/intel-me.json $(sbom-me-bin) | $(build-dir) $(sbom-me-nftp-tool)
 	cp $< $@
-	set -e; \
-	me_ver=$$(strings -a "$(CONFIG_ME_BIN_PATH)" \
+	me='$(sbom-me-bin)'; \
+	: Try to extract from string first. Does not always work; \
```
miczyg1: What are these lines starting with `:` ?

src/sbom/Makefile.mk
```diff
@@ -170,22 +179,83 @@ $(build-dir)/compiler-%.json: $(src-dir)/compiler-%.json | $(build-dir)/goswid
...
+    sbom-me-nftp-tool := $(IFWITOOL)
+else
+    sbom-me-bin := $(CONFIG_ME_BIN_PATH)
+    ifeq ($(CONFIG_SOC_INTEL_CSE_HAVE_SPEC_SUPPORT),y)
+        sbom-me-nftp-tool := $(CSE_SERGER)
+        sbom-me-bpdt-version := $(if $(filter 12 13 14,$(CONFIG_ME_SPEC)),1.6,1.7)
+    else
+        sbom-me-nftp-tool := $(CSE_FPT)
+    endif
+endif
```
miczyg1, line 214: Not sure why we set these `sbom-me-nftp-tool`. ifwitool, cse_serger and cse_fpt are always built I think... So We can just list them in the dependencies below

src/sbom/Makefile.mk
```diff
@@ -170,22 +179,83 @@ $(build-dir)/compiler-%.json: $(src-dir)/compiler-%.json | $(build-dir)/goswid
...
+# 3. CONFIG_SOC_INTEL_CSE_HAVE_SPEC_SUPPORT=n
+#    => ME_SPEC <= 11, no BPDT, use cse_fpt to extract NFTP
+ifeq ($(CONFIG_NEED_IFWI),y)
+    sbom-me-bin := $(call strip_quotes,$(CONFIG_IFWI_FILE_NAME))
+    sbom-me-nftp-tool := $(IFWITOOL)
+else
+    sbom-me-bin := $(CONFIG_ME_BIN_PATH)
+    ifeq ($(CONFIG_SOC_INTEL_CSE_HAVE_SPEC_SUPPORT),y)
+        sbom-me-nftp-tool := $(CSE_SERGER)
+        sbom-me-bpdt-version := $(if $(filter 12 13 14,$(CONFIG_ME_SPEC)),1.6,1.7)
```
miczyg1: Wouldn't it be easier to do it liek this:

1. src/soc/intel/common/block/cse/Kconfig:
```
config CSE_BPDT_VERSION
	string
	default "1.6" if SOC_INTEL_CSE_HAVE_SPEC_SUPPORT && (ME_SPEC < 15)
	default "1.7" if SOC_INTEL_CSE_HAVE_SPEC_SUPPORT && (ME_SPEC >= 15)
	help
	  This config indicates the BPDT version used by CSE for a given SoC.
```

Then here we could simply use CSE_BPDT_VERSION

miczyg1 (changes requested):

src/sbom/Makefile.mk
philipanda: It's not possible to use `#` in such one line script. I've used colon, which is a noop in bash, to have a way to embed a comment inside the script instead of in one block before the recipe.

Going into technicalities, It works like if the actual comment words are arguments to the `:` command, which does nothing and succeeds, similarly to `true`. So this is not a real comment, the text is expanded as args, then discarded.

src/sbom/Makefile.mk
philipanda: Sure, why not. This information would then be available in the `.config` embedded in the binary, so maybe it would allowe to reduce some potential redundancy if other tools or scripts would need to find out this BPDT version.

https://github.com/Dasharo/coreboot/pull/941/changes/e1fda6e4ac752a0f430dcc264293a74613abf0bd

src/sbom/Makefile.mk
philipanda: Sure https://github.com/Dasharo/coreboot/pull/941/changes/83a9bab268bd64ec232d63eb3a394ef659e99ac9

src/sbom/Makefile.mk
miczyg1: I see. Well, I haven't seen inline comments anywhere in Makefiles. Comments are usually either omitted, or present before a function or formula. I don't think we need that deep comments. IF anything, I would put it above the formula/function, if something needs more explanation

philipanda: https://github.com/Dasharo/coreboot/pull/941/commits/1302c978009b95351434d21c8171e46925190739

miczyg1 (approved):
All commits in this PR have Upstream-Status so approving

PR #943, opened 2026-07-20
amdblocks/spi: Fix SPI ROM mapping

miczyg1:
The CB:86620 change introduced checks for SPI ROM mapping during BIOS mmap windows initialization. This caused bricks on certain systems, because the fch_spi_rom_remapping does not give a complete view of how SPI ROM is mapped via ROM2 and ROM3.

The SPI_ROM_PAGE register contains the final XOR value applied to the host address to obtain the flash offset. However, for ROM2 there is also the SPI_ROM2_OVERRIDE register, which can apply bit masks on the ROM2 MMIO address bits 25:24 and forcefully remap host address to different flash offsets, or so called pages/banks. ROM3 is not affected by SPI_ROM2_OVERRIDE, only SPI_ROM_PAGE applies. Except when the bit 4 of SPI_ROM2_OVERRIDE is set, which enforces every 16MiB chunk in ROM3 to map to the SPI ROM bank 3 (for 64MB flash, it would be offset 0x3000000).

To get a final ROM2 mapping configuration for SPI ROM, the SPI_ROM2_OVERRIDE settings have to be applied first, then XORed with SPI_ROM_PAGE content.

Also different rules apply to ROM2/ROM3 if SPI ROM is configured for 32bit and 24bit address. With 24bit SPI ROM, the XOR does not apply.

Additionally for ROM3 mapping, allow a single 16MiB chunk to be mapped regardless if ROM3 mapping is linear or not. We only force ROM2 to map to flash offset 0, then the remaining 16MiB chunk of 32MiB flash is mapped at the ROM3 bank configured by FCH SPI registers. In the case flash is bigger than 32MiB, we require the ROM3 to be linear.

The implementation has been confirmed to work on the following SoCs:

- Phoenix AM5: 32MiB flash, SPI_ROM2_OVERRIDE = 0x14cc, SPI_ROM_PAGE = 0x00
- Turin: 32MiB flash, SPI_ROM2_OVERRIDE = 0x14c0, SPI_ROM_PAGE = 0x03

Information about SPI ROM mapping can be found in Phoenix A2 step PPR (doc 68053) and Turin PPR (doc 57238) section: "Programming for ROM Protection register". Additional information about FCH SPI ROM mapping is also available in Turin PPR in section: "ROM Address Mapping Support in FCH implementation".

TEST=Start bootblock on Gigabyte MZ33-AR1 (Turin) and MSI PRO B850-P WIFI (Phoenix AM5) and see that the die function on non-default mapping is no longer executed. SOC_AMD_COMMON_BLOCK_SPI_MMAP untested.

Change-Id: I215b63d1bc5626af62522f4b7daf4f779fa98bb7

Upstream-Status: Backport [CB:94146]

### DIFF
--- a/src/soc/amd/common/block/include/amdblocks/spi.h
+++ b/src/soc/amd/common/block/include/amdblocks/spi.h
@@ -76,6 +76,17 @@ enum spi100_speed {
 #define SPI100_HOST_PREF_CONFIG		0x2c
 #define   SPI_RD4DW_EN_HOST		BIT(15)
 
+#define SPI_ROM2_OVERRIDE		0x30
+#define   SPI_ROM2_ADDR_BIT24_VAL	BIT(0)
+#define   SPI_ROM2_ADDR_BIT25_VAL	BIT(1)
+#define   SPI_ROM2_ADDR_BIT24_MASK	BIT(2)
+#define   SPI_ROM2_ADDR_BIT25_MASK	BIT(3)
+/* Below bit is only mentioned in PPRs in section
+ * "Programming for ROM Protection register".
+ * The register definition marks this bit as reserved.
+ */
+#define   SPI_FORCE_ROM3_MAP_TO_BANK3	BIT(4)
+
 #define SPI_STATUS			0x4c
 #define   SPI_DONE_BYTE_COUNT_SHIFT	0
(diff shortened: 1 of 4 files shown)


mkopec (approved):
Tested: ASRock TURIND8UD still boots with this change.

PR #944, opened 2026-07-21
capsule.sh: Keep capsule flags when resigning capsules

miczyg1:
The resign subcommand did not preserve original capsule flags because coreboot config was not sourced in that path. Try to extract the config file from the inner capsule payload binary, which should be the coreboot.rom file.

Upstream-Status: Inappropriate [custom script]
Change-Id: I50997ba01220d8866b238bfff0be69548f4d0aa9

### DIFF
--- a/capsule.sh
+++ b/capsule.sh
@@ -71,8 +71,10 @@ function help_subcommand() {
 }
 
 function source_coreboot_config() {
-    if [ ! -f .config ]; then
-        die "no '.config' file in current directory"
+    local config_file=${1-.config}
+
+    if [ ! -f "$config_file" ]; then
+            die "config file '$config_file' doesn't exist"
     fi
 
     local line
@@ -81,7 +83,7 @@ function source_coreboot_config() {
         if ! eval "$line"; then
             die "failed to source '.config'"
         fi
-    done <<< "$(sed 's/\$(\([^)]\+\))/${\1}/g' .config)"
+    done <<< "$(sed 's/\$(\([^)]\+\))/${\1}/g' $config_file)"
 }
 
 function require_capsule_support() {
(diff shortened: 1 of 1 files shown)


capsule.sh
```diff
@@ -71,8 +71,18 @@ function help_subcommand() {
...
+    if [ $# -ne 1 ]; then
+        if [ ! -f .config ]; then
+            die "no '.config' file in current directory"
+        fi
+        config_file=.config
+    else
+        if [ ! -f $1 ]; then
+            die "no '.config' file in current directory"
+        fi
+        config_file=$1
```
SergiiDmytruk:
```suggestion
    local config_file=${1-.config}

    if [ ! -f "$config_file" ]; then
            die "config file '$config_file' doesn't exist"
    fi
```

capsule.sh
```diff
@@ -507,6 +518,13 @@ function resign_subcommand() {
     metadata[sign_cert]=$sign_cert
     metadata[sub_cert]=$sub_cert
     metadata[root_cert]=$root_cert
+
+    local -A metadata_inner
+    mkdir -p "$tmp_dir/inner" || exit 1
```
SergiiDmytruk:
```suggestion
    mkdir "$tmp_dir/inner"
```
There is `set -e` at the top and `$tmp_dir` must already exist at this point.

SergiiDmytruk (approved):
Just code improvements.

capsule.sh
miczyg1: https://github.com/Dasharo/coreboot/compare/82a8bfc96c7be92b757c24166a0c2bfb1476d6f6..0c15e9f6a843628ee7f03ba123178b901095c597

capsule.sh
miczyg1: https://github.com/Dasharo/coreboot/compare/82a8bfc96c7be92b757c24166a0c2bfb1476d6f6..0c15e9f6a843628ee7f03ba123178b901095c597

PR #947, opened 2026-07-24
SBOM further issue fixes

filipleple:
*ref: sha-1717*

### DIFF
--- a/src/sbom/Makefile.mk
+++ b/src/sbom/Makefile.mk
@@ -248,11 +248,25 @@ coreboot-gitdir := $(shell git rev-parse --git-dir)
 # colloquial_version is the latest coreboot release, which is why this is the one
 # component not using sbom-git-latest-rel: forks carry their own release tags
 # (e.g. novacustom_nuc_box_v0.9.2), and git describe would return those instead
-# of the coreboot release. Assumes static release tag format, might break.
+# of the coreboot release. Releases are matched by tag name rather than by tag
+# message ("coreboot version <X.Y>"), the less stable of the two: the name has
+# kept its <X.Y>[.<Z>] shape since 4.0, while the message wording changed to
+# "coreboot release <X.Y>" in 26.03. The optional third component covers point
+# releases (24.02.01, 4.20.1); those only exist on release branches, so they are
+# reachable only from a fork based on one.
+#
+# An empty result means the checkout carries no release tag at all, i.e. a
+# shallow or blobless clone made without --tags. That is an error rather than an
+# omitted field, unlike the tagless third-party repositories handled by
+# sbom-git-latest-rel: for coreboot itself the release is always there to find.
 $(build-dir)/coreboot.json: $(src-dir)/coreboot.json $(coreboot-gitdir)/HEAD | $(build-dir)/goswid
 	cp $< $@; \
 	git_comm_hash=$$($(call sbom-git-comm-hash,.)); \
-	git_latest_rel=$$(git tag --merged HEAD | grep -E '^[0-9]{1,2}\.[0-9]{1,2}$$' | sort -V | tail -n1); \
+	git_latest_rel=$$(git tag --merged HEAD | grep -E '^[0-9]{1,2}\.[0-9]{1,2}(\.[0-9]{1,2})?$$' | sort -V | tail -n1); \
+	if [ -z "$$git_latest_rel" ]; then \
+		echo "ERROR: no coreboot release tag reachable from HEAD, cannot set colloquial-version; run 'git fetch --tags'" >&2; \
+		exit 1; \
+	fi; \
 	sed -i -e "s/<software_version>/$$git_comm_hash/" \
 		-e "s/<colloquial_version>/$$git_latest_rel/" \
 		$@;\
--- a/src/sbom/Makefile.mk
+++ b/src/sbom/Makefile.mk
@@ -468,9 +468,10 @@ $(build-dir)/amd-agesa.json: $(src-dir)/amd-agesa.json $(CONFIG_AGESA_BINARY_PI_
 		sed -i "/software-version/d" $@; \
 	fi
 
-# Record the git commit of the openSIL source tree as the SBOM version.
+# Record the git commit of the openSIL source tree as the SBOM version, in the
+# human-readable "<commit-date>_<hash>" form used for the other git-backed deps.
 # openSIL is an open-source library checked out as a submodule or external
-# repo, so the commit hash is the canonical version identifier.
+# repo, so the commit is the canonical version identifier.
 $(build-dir)/amd-opensil.json: $(src-dir)/amd-opensil.json | $(build-dir)
 	cp $< $@
 	set -e; \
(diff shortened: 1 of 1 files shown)


src/sbom/Makefile.mk
```diff
@@ -593,9 +595,9 @@ $(build-dir)/payload-edk2-platforms.json: $(src-dir)/payload-edk2-platforms.json
 	cp $< $@
 	set -e; \
 	if [ -e "$(edk2-platforms-git-dir)/.git" ]; then \
-		git_tree_hash=$$(git --git-dir $(edk2-platforms-git-dir)/.git log -n 1 --format=%T); \
-		git_comm_hash=$$(git --git-dir $(edk2-platforms-git-dir)/.git log -n 1 --format=%H); \
-		sed -i -e "s/<colloquial_version>/$$git_tree_hash/" -e "s/<software_version>/$$git_comm_hash/" $@; \
+		git_comm_hash=$$(git --git-dir $(edk2-platforms-git-dir)/.git log -n 1 --format="%cs_%H"); \
+		git_latest_rel=$$(git --git-dir $(edk2-platforms-git-dir)/.git tag --merged HEAD --sort=-creatordate | head -n1); \
```
SergiiDmytruk: This may be a reimplementation of `git describe --abbrev=0` assuming `--sort=-creatordate` doesn't interfere (should make no difference when there is a tag closer than any merge commit).

This is just a note, since the outcome should be the same.  But we could consider changing the command at some point (e.g., when upstreaming, preferably after deduplicating all the places at which it's used).

SergiiDmytruk (dismissed):

src/sbom/Makefile.mk
filipleple: https://github.com/Dasharo/coreboot/pull/947/changes/950787fd02b8ee0cb785a4c45f6c0ad5f0eb6e95

src/sbom/Makefile.mk
```diff
@@ -220,10 +245,13 @@ $(build-dir)/compiler-%.json: $(src-dir)/compiler-%.json | $(build-dir)/goswid
 
 coreboot-gitdir := $(shell git rev-parse --git-dir)
 # coreboot version, format like "<commit-date>_<hash>"
-# colloquial_version is latest coreboot release. Assumes static release tag format, might break.
+# colloquial_version is the latest coreboot release, which is why this is the one
+# component not using sbom-git-latest-rel: forks carry their own release tags
+# (e.g. novacustom_nuc_box_v0.9.2), and git describe would return those instead
+# of the coreboot release. Assumes static release tag format, might break.
```
philipanda: I see that coreboot release tags always:
a) are signed
b) contain `coreboot version <X.Y>` in the description.

Maybe these details could be used to help with it potentially breaking in the future?
We could do a shallow clone of the upstream and get the list of freshest tags from there. We could then see which is the newest one that exists on the downstream.

filipleple: Both hold from 4.1 to 25.12, but `26.03` says `coreboot release`, not `coreboot version`, so the message wording is actually less stable than the tag name scheme, and signedness doesn't discriminate since `novacustom_nuc_box_v0.9.2` is itself a signed annotated tag. Fetching upstream might also make the same commit produce different SBOM content depending on when and where it's built, which I'd rather avoid in a CRA artifact, so I'll keep the name regex and just widen it for point releases plus fail loudly when it comes back empty.

https://github.com/Dasharo/coreboot/pull/947/commits/52aa9a5db12bae485df310089497ebb4d068e9e9

philipanda: > but 26.03 says coreboot release, not coreboot version

hmm, right

> Fetching upstream might also make the same commit produce different SBOM

I thought fetching the upstream could be used just for the sake of finding out what tag name is the latest, because it doesn't contain noise like `novacustom_nuc_box_v0.9.2`. 

> fail loudly when it comes back empty.

Would it be possible to fail? I think only if you had removed all the tags from the repo. I think failing loudly in some way would be good in these sbom makefiles, I'm not sure about breaking the whole build, because if there are N errors then one would need to build the firmware N times to find out about all of them. Maybe it would be good to insert the error message into some sbom field and only then fail? Or gather the fail messages and only fail the build on the end? I guess that would be unnatural to do in makefile.

I'm throwing some ideas, but don't feel obliged to address them.

philipanda (dismissed):

src/sbom/Makefile.mk
```diff
@@ -220,11 +245,28 @@ $(build-dir)/compiler-%.json: $(src-dir)/compiler-%.json | $(build-dir)/goswid
...
+# message ("coreboot version <X.Y>"), the less stable of the two: the name has
+# kept its <X.Y>[.<Z>] shape since 4.0, while the message wording changed to
+# "coreboot release <X.Y>" in 26.03. The optional third component covers point
+# releases (24.02.01, 4.20.1); those only exist on release branches, so they are
+# reachable only from a fork based on one.
+#
+# An empty result means the checkout carries no release tag at all, i.e. a
+# shallow or blobless clone made without --tags. That is an error rather than an
+# omitted field, unlike the tagless third-party repositories handled by
+# sbom-git-latest-rel: for coreboot itself the release is always there to find.
```
SergiiDmytruk, line 261: > unlike the tagless third-party repositories handled by sbom-git-latest-rel

This doesn't sound right, `sbom-git-latest-rel` is used to find tags.

SergiiDmytruk (dismissed):

src/sbom/Makefile.mk
philipanda: I think it was supposed to mean that `sbom-git-latest-rel` also handles the situation when there are no tags, but I don't think it's the case. It just prints nothing if there are no tags. That's not handling, that's ignoring the issue.
On the other hand it might have meant that it's the only place where we verify the tag exists, because we know coreboot has tags. Failing if theres no tag in repos with no tags wouldn't be right. But just like coreboot, we know whether tags exist or not on all of the repos used here, so why not just make it explicit.

I agree this sentence could be better.

filipleple: https://github.com/Dasharo/coreboot/pull/947/commits/b67d7a133d0296fd58c689e65b6e42bbc1c01cef

SergiiDmytruk (approved):

SergiiDmytruk:
Rebased on top of the current `dasharo`: https://github.com/Dasharo/coreboot/compare/b67d7a133d0296fd58c689e65b6e42bbc1c01cef..e06e05495939e4bfb34cad850b83cda0f5a229e0

SergiiDmytruk:
`test_q35` fails trying to clone `osfv-scripts` from git.3mdeb.com.  @filipleple, was [this change](https://github.com/Dasharo/open-source-firmware-validation/commit/c053d17f8bcc5157850af956a146de59c467cd95#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552) committed accidentally?  It points at a commit which is https://github.com/Dasharo/osfv-scripts/commit/85ea43153ef8f032ed9c86e4be81a2a49736d0e2 before being rebased (apparently, automatically).  If this wasn't intended, then it's not to be fixed in here and is something for addressing in OSFV.

filipleple:
@SergiiDmytruk yes that was a bad change, sorry, got a bit mixed up in the gitea-github workflow. Fixing it here: https://git.3mdeb.com/dasharo/open-source-firmware-validation/pulls/123

PR #948, opened 2026-07-24
SBOM b850 blobs

filipleple:
Mocked up; we don't know yet where the blobs will land and how they're structured, i.e. how to obtain the version. Here is the template for including them in the SBOM, calculating the hash, and so far filling the version with containing repo's commit hash.


miczyg1:
@filipleple thanks. I took this over and added missing pieces.

PR #949, opened 2026-07-24
Tgl dgpu detect

philipanda:


mkopec:
we should probably also add the same to the V5x0TNx mainboard code, to avoid the menu disappearing after the edk2 changes. EDK2 now expects the variable to exist if dgpu option is supposed to be shown

payloads/external/edk2/Kconfig.dasharo
```diff
@@ -4,7 +4,7 @@ config EDK2_REPOSITORY
 	default "https://github.com/Dasharo/edk2"
 
 config EDK2_TAG_OR_REV
-	default "2d3fe517ad3a608741f14a37d9fbf8529060ae26"
+	default "0511646f4ee4c696ff4a247f74cef04cad7da602"
```
mkopec:
```suggestion
	default "93bc0155479d2e153264ffcd8be37fc2847490a1"
```

I resigned your edk2 commit, the hash changed

philipanda:
> we should probably also add the same to the V5x0TNx mainboard code, to avoid the menu disappearing after the edk2 changes. EDK2 now expects the variable to exist if dgpu option is supposed to be shown

I think it should work no matter the platform. The default implementation always says the dGPU is available. It can be overwritten to actually allow it to be set to false. The `set_dgpu_present_option` is registered to always run in `src/vendorcode/dasharo/options.c` (BOOT_STATE_INIT_ENTRY). The DGPUPresent should always be set to 1 on any other platform that enables the dGPU hybrid graphics mode menu option.

mkopec (changes requested):

mkopec:
ah, i see now.

payloads/external/edk2/Kconfig.dasharo
philipanda: https://github.com/Dasharo/coreboot/compare/5d1887509316b6c5952a0460d13f74ee73608c19..1ab4e1cbc58a842dc860ae8c1b4e5189632fc9bf

mkopec (approved):

PR #37, opened 2021-11-10
WIP: Northbridge fixes

miczyg1:

PR #957, opened 2026-08-14
Ap2110 unpro

filipleple:


filipleple:
Trying to mark as ready, but the functionality seems down at the moment:

image

configs/config.protectli_ap2110
```diff
@@ -9,7 +9,7 @@ CONFIG_SBOM_IFD=y
 CONFIG_SBOM_VBOOT=y
 CONFIG_SBOM_IPXE=y
 CONFIG_VENDOR_PROTECTLI=y
-CONFIG_VBOOT=y
+CONFIG_FMDFILE="src/mainboard/$(CONFIG_MAINBOARD_DIR)/vboot-rwa.fmd"
```
SergiiDmytruk: Did you intend to keep `CONFIG_SBOM_VBOOT`?

src/mainboard/protectli/vault_adl_n/board.fmd
```diff
@@ -0,0 +1,17 @@
+FLASH@0xff000000 16M {
+	SI_ALL {
+		SI_DESC 4K
+		SI_ME
+	}
+	SI_BIOS@0x500000 0xb00000 {
+		SMMSTORE(PRESERVE) 256K
```
SergiiDmytruk, line 7: Do we want to allocate 512K for SMMSTORE to not have to grow it later as was on other platforms?

SergiiDmytruk (commented):
The PR needs to be resend (but probably not rebased after reaching RC3), providing review here for now.

configs/config.protectli_ap2110
filipleple: Hmm, I was in two minds about this, because technically the vboot lib does still get compiled in due to measured boot, so it makes some sense to track its version. But, I think this can be considered a reach and would most likely create more confusion than value, so it's now disabled: [new PR diff](https://git.3mdeb.com/dasharo/coreboot/pulls/37/files#diff-f41147a2fac012983254fcf9b9e277b0f7f22ae5)

src/mainboard/protectli/vault_adl_n/board.fmd
filipleple: Yes, that's a good call, [added](https://git.3mdeb.com/dasharo/coreboot/pulls/37/files#diff-73ad706377d91baeba2ca7c42d1db062df17f760)

filipleple:
Sent internally as [#37](https://git.3mdeb.com/dasharo/coreboot/pulls/37)

